PR #38, opened 2020-02-20
Document how to run tests

### LINKED ISSUE #37: Document how to run tests locally
To help people get involved.

avishayp:
Fix #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ To run tests locally pip-install `tox` and run it:
 ```
 tox
 ```
+`make` does the same thing.
 
 We recommend using pipenv for a virtual environment:
 
--- a/README.md
+++ b/README.md
@@ -24,35 +24,35 @@ The codebase has some momentum already, has been continuously developed since Se
 ## Overview of the Project
 Planned efforts include:
 
-+ **Combined Model Implementation**  
++ **Combined Model Implementation**
 Having copies of the underlying model in ~80 Excel files is no longer tenable, the level of toil involved in making substantial changes to the model is prohibitive. The project intends to deliver a single implementation in Python which can model the current ~80 solutions developed by Project Drawdown and allow easier development of more solution models in the future.
-Current status:  
+Current status:
     + Python currently implements 70 out of the 80 Excel solution models of the 2019 version. The design doc for this effort was: [Drawdown Software Remodeling: backend models](https://docs.google.com/document/d/1X9X-61CG26m0XTUmqKeGJwU-HinPELD9HwBO064b5dA/edit) (now considered complete).
     + The remaining 10 of the 80 are solutions which differ more extensively from the rest, like Food Waste and Family Planning. Python implementations of the remaining 10 solutions have not been started. There is nothing blocking effort on these, just lack of cycles.
     + Updates to the Python code to support the 2020 versions of the 70 solutions has begun, and several solutions from the 2020 Excel update are working.
-    + Automated testing is a key goal of the Python effort, to enable future work on the model to proceed with confidence. To ensure the new implementation faithfully reproduces the original, there is a test which runs the Excel any Python models for a given solution and checks that their results match within a floating point margin of error at every step of the model calculation. There is a [YouTube video describing the levels](https://www.youtube.com/watch?v=K6P56qUkCrw) of automated testing.  
+    + Automated testing is a key goal of the Python effort, to enable future work on the model to proceed with confidence. To ensure the new implementation faithfully reproduces the original, there is a test which runs the Excel any Python models for a given solution and checks that their results match within a floating point margin of error at every step of the model calculation. There is a [YouTube video describing the levels](https://www.youtube.com/watch?v=K6P56qUkCrw) of automated testing.
     [![codecov](https://codecov.io/gh/ProjectDrawdown/solutions/branch/master/graph/badge.svg)](https://codecov.io/gh/ProjectDrawdown/solutions)
 
-+ **UI aimed at researchers**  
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -24,35 +24,35 @@ The codebase has some momentum already, has been continuously developed since Se
 ## Overview of the Project
 Planned efforts include:
 
-+ **Combined Model Implementation**
```
DentonGentry: Ah, Github markdown doesn't need double spaces at the end to add a line-break. Cool. I put those in out of habit from a different environment.

README.md
```diff
@@ -68,6 +68,11 @@ $ git clone https://github.com/ProjectDrawdown/solutions.git
 $ cd solutions
 ```
 
+To run tests locally pip-install `tox` and run it:
+```
+tox
```
DentonGentry: Hmm, tox is interesting. I've been using make, and to be perfectly honest hadn't considered that there exist people who don't habitually use Makefiles for everything. I'm happy to include tox, though I may come back to edit this README again to add mention of the Makefile as well.

One thing which would be good in the tox support: tests which take a really long time to run carry the attribute pytest.mark.slow. So for example, "make test" runs tests which are not marked slow and completes in under 2 minutes, while "make alltest" runs everything and takes over an hour.

I browsed tox.readthedocs.io but didn't immediately see a way to have multiple targets like this, are you aware of one?

The Makefile also passes --strict-markers to pytest, as otherwise it is too easy to have a typo in pytest.mark attributes and accidentally cause a test to always be skipped. pytest.ini lists the valid marks.

avishayp: Sorry missed Makefile! I'm so used to having a top-level `tox.ini` in python-focused code base. The 👍 for `tox` is it creates the virtual env itself. I will add support for all relevant Makefile sections and have the Makefile call tox. Does that work?

avishayp:
Now `make` and `make alltest` call `tox` with same cli args as previously specified in the Makefile. The current target python version is 3.7 (pretty arbitrary). Should it be 3.8?

DentonGentry:
I think Python 3.7 is fine as the minimum version.

DentonGentry:
I had to (hopefully temporarily) revert some of this because tox isn't working for me:

```
(solutions) bash-3.2$ tox
py38 create: /path/to/solutions/.tox
ERROR: cowardly refusing to delete `envdir` (it does not look like a virtualenv): /path/to/solutions/.tox
_______________________________ summary _______________________________
ERROR:   py38: error
(solutions) bash-3.2$
```

I considered reverting locally, but didn't especially want to have the instructions say one thing but me do another. I do like the idea of using tox and intend to keep trying to get it working here, and then restore it.

I also updated tox to use Python 3.8 because pipenv has 3.8 installed, to avoid consuming disk space for a parallel 3.7 installation. I left the tox.ini file, hopefully you can directly invoke tox and have it still work. I only reverted the Makefile to call pytest again.

avishayp:
I'd guess we're not using same `tox` version.
```
$ which tox
/usr/local/bin/tox

$ tox --version
3.5.3 imported from /usr/local/lib/python3.7/site-packages/tox/__init__.py
```

DentonGentry:
(solutions) bash-3.2$ which tox
/Users/dgentry/.local/share/virtualenvs/solutions-4QOqvuZn/bin/tox
(solutions) bash-3.2$ tox --version
3.14.5 imported from /Users/dgentry/.local/share/virtualenvs/solutions-4QOqvuZn/lib/python3.8/site-packages/tox/__init__.py
(solutions) bash-3.2$

I instrumented the code near "ERROR: cowardly refusing to delete `envdir`":

```
707     if not (
708         # doesn't exist => OK
709         not venv.path.exists()
710         # does exist, but it's empty => OK
711         or not dir_items
712         # tox has marked this as an environment it has created in the past
713         or ".tox-config1" in dir_items
714         # it exists and we're on windows with Lib and Scripts => OK
715         or (INFO.IS_WIN and dir_items > {"Scripts", "Lib"})
716         # non-windows, with lib and bin => OK
717         or dir_items > {"bin", "lib"}
718         # pypy has a different lib folder => OK
719         or dir_items > {"bin", "lib_pypy"}
720     ):
721         venv.status = "error"
723         reporter.error(
724             "cowardly refusing to delete `envdir` (it does not look like a virtualenv): "
725             "{}".format(venv.path)
726         )
```

The one which is triggering the error is the "not dir_items" check. When the .tox directory is created, on my system a ".tmp" file appears within it and this causes tox to get the error because the directory is not empty.

I haven't figured out if tox itself is creating the .tmp file, or it is something else in my environment.

DentonGentry:
If I add to that conditional that ".tmp" is ok (line 720):

```
718         # pypy has a different lib folder => OK
719         or dir_items > {"bin", "lib_pypy"}
720         or dir_items == {".tmp"}
721     ):
```

tox proceeds to create the venv and run.

tox does create a {toxworkdir}/.tmp directory, I think the directory I'm seeing is the one created by tox. I don't understand what is different about my environment, clearly most people do not see this problem.

DentonGentry:
The presence of the envdir line in tox.ini is part of the issue. If I remove it, tox runs successfully. If I get tox to run at least once, I can put the envdir line back and it will continue to work.

I suspect the .tox directory was already present when you added envdir to tox.ini, and so you never ran into this.

```
[testenv]
# one virtual env for all test - without this line tox will install .tox/all env to run the 'all' suit
envdir = {toxinidir}/.tox
```

avishayp:
`pip3 install tox==3.5.3` will solve it.

PR #44, opened 2020-02-21
Make tox the main test runner

### LINKED ISSUE #11: pytest to show DeprecationWarnings in our own code
We began running pytest -W ignore::DeprecationWarning, as hundreds of deprecation warnings from site-packages were overwhelming. Unfortunately this has concealed a number of deprecation warnings in code which we do actively develop, such as things which break in [Pandas 1.0 (see issue #4)](https://github.com/ProjectDrawdown/solutions/issues/4).

This seems like a problem which should have a solution, to restrict pytest warnings to local code and ignore site-packages.

avishayp:
Fix #11 
continuing #38 

### What this PR does

 * Fix tox issue by changing `envdir` to something more sensible
 * Make `tox` the single point test runner - it's called from Makefile and github workflow
 * Remove `pytest.ini` and move its content to `tox.ini`
 * Use `once::DeprecationWarning` instead of `ignore::DeprecationWarning`

### DIFF
--- a/.github/workflows/drawdown.yml
+++ b/.github/workflows/drawdown.yml
@@ -20,7 +20,7 @@ jobs:
         python -m pip install --upgrade pip
     - name: Test with tox
       run: |
-        pip install tox==3.14.0
+        pip install tox
         tox -e ci
     - name: Export to codecov.io
       uses: codecov/codecov-action@v1.0.2
--- a/.github/workflows/drawdown.yml
+++ b/.github/workflows/drawdown.yml
@@ -18,11 +18,10 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
-    - name: Test with pytest
+    - name: Test with tox
       run: |
-        pip install pytest
-        pytest -v --cov=. --cov-report term --cov-report=xml --strict-markers
+        pip install tox==3.14.0
+        tox -e ci
     - name: Export to codecov.io
       uses: codecov/codecov-action@v1.0.2
       with:
--- a/tox.ini
+++ b/tox.ini
@@ -1,21 +1,28 @@
 [tox]
-envlist=py38
-skipsdist=True
+envlist = py38
+skipsdist = True
 
 [testenv]
 # one virtual env for all test - without this line tox will install .tox/all env to run the 'all' suit
-envdir = {toxinidir}/.tox
+envdir = {toxworkdir}/common
 
 deps =
     -rrequirements.txt
 
 commands =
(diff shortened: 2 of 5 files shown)


tox.ini
```diff
@@ -1,21 +1,28 @@
...
 
-[testenv:all]
+# runs on github workflow
+[testenv:ci]
 commands =
-    pytest -v --strict-markers -W ignore::DeprecationWarning --cov=. --cov-report term-missing {posargs}
+    pytest --cov=. --cov-report term --cov-report=xml
+
+[pytest]
+addopts = -v -W once::DeprecationWarning --strict-markers
```
DentonGentry: I'd request to make this ignore::DeprecationWarning again, as discussed on issue #11 I don't think once is really preferable.

.github/workflows/drawdown.yml
```diff
@@ -18,11 +18,10 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -r requirements.txt
-    - name: Test with pytest
+    - name: Test with tox
       run: |
-        pip install pytest
-        pytest -v --cov=. --cov-report term --cov-report=xml --strict-markers
+        pip install tox==3.14.0
```
DentonGentry: Could we leave the version unspecified, or make it >=3.14.0? I understand this means some future tox update might make the workflow start to fail, but I think I'd prefer that rather than find we're still using 3.14.0 years from now. I'd prefer to let it update and deal with future breakage if it happens.

tox.ini
```diff
@@ -1,21 +1,28 @@
...
+ignore = .tox .git data
```
DentonGentry: adding .tox and .git to the ignore list is fine, but I'd prefer not to add data. There may not be any tests in there now, but I could see adding data sanity tests later and silently ignoring them wouldn't be good.

DentonGentry:
As one of the requests is to not change to once::, then the PR title and description would need to be updated as well.

avishayp:
@DentonGentry thanks, changed accordingly. So this PR is only about using `tox`. I've so far failed to implement `filterwarnings` to achieve a proper fix to #11 

But I think this PR is useful as-is becaue if/when I do fix that it'll be only in `tox.ini`.

DentonGentry:
https://github.com/ProjectDrawdown/solutions/blob/master/.zenodo.json is used to generate an acknowledgement list for papers published which use results generated by this software. I'm expecting to additionally check git history to acknowledge people in any papers I'm involved in, but I'm not always involved. If whoever is publishing uses zenodo.com (which is pretty widespread in academic publishing), zenodo.json would still be processed.

So: if you're interested, I'd recommend sending a PR to add yourself to zenodo.json.

PR #46, opened 2020-02-22
Move status badges to top of README

avishayp:
Add workflow status, put it all on top of readme.


avishayp:
This should be my last unsubstantial PR.

Issue #47, opened 2020-02-22
Static code analysis


avishayp:
Suggestion:

1.  Do a pylint-driven code cleaning
2.  Add `pylint` to the workflow

I've started with [1], but I already see that fixing the issues raised by `pylint` requires a deeper level of familiarity with the codebase. So some help will be needed.

If that's ok, I'll submit a PR to clean unused vars, unused imports, redundant semicolons etc.

DentonGentry:
Sure, pylint would be great. I wish we'd done it sooner, just didn't get it done.

We started with the Google Python coding style (2 space indent) and later reformatted to be closer to PEP8. There may be a few files left which are 2 space indent, and can be reformatted to 4 space.

avishayp:
Let's handle inconsistent indent in separate PR it'll create a huge difficult to review diff for you.

avishayp:
To update: am working on it. Analysis is not reliable enough to be blindly trusted, I've broken some tests removing imports or variables marked as unused by pylint. Which also makes me reconsider adding pylint to the CI workflow.

DentonGentry:
Certainly, every time I've tried to use pylint on nontrivial amounts of code it has involved #pylint comments to resolve occasional inexplicable behaviors. Nonetheless I've found pylint to be valuable for finding real problems which otherwise would not have been noticed until much later, I'd encourage the effort spent on it.

tpltnt:
Is issue is very broad. Maybe it should be split into smaller, well-defined PRs with a definition of done for each. (i.e. one for indentation, one for another class of issues, one for adding pylint explicity to the CI pipeline)

avishayp:
The DoD of this issue:

 * Ad-hoc discovery and fix of some low-hanging bugs (arguably addressed in #53 )
 * Converge to an agreed-upon set of pylint W/E/F flags which can be put into workflow without creating too much noise (or decide against that)

The bikeshedding part is under #48

avishayp:
By stated DoD #61 should close this.

Issue #48, opened 2020-02-23
Fix inconsistent indentation


avishayp:
pylint tells me the autogenerated files are mostly using 2-space ident.
This issue is to reformat all files to 4-space indentation.

avishayp:
@DentonGentry considering this will be practically unreviewable, is there anything I could could test on top of the test suite to be confident the reformat didn't break anything?

DentonGentry:
I'd recommend to run the Jupyter notebook, we don't (yet) have automated tests that the notebook still works.

I'd request not to reformat the autogenerated files in solution/*/__init__.py, as we're very soon (within days) going to begin importing new versions of the solutions which would generate new files with two space indent. If you fix the rest of the tree, I'll redo tools/solution_xls_extract.py to use four space indent before we begin importing the new files.

jonathan-s:
May I suggest the library black for formatting? https://github.com/psf/black

DentonGentry:
While I understand the purpose of something like https://github.com/psf/black, in the same general idea as gofmt for eliminating bikeshedding about whitespace, I'm not all that enthused about adopting something like that just yet as the arbiter of code style.

PR #53, opened 2020-02-27
Lint

avishayp:
### in this PR
Very partial fix for #47 

I'm submitting this PR for review with not much confidence it should be merged, so we could discuss #47 based on facts and decide how/if to proceed.

The good news is that 2-3 obvious bugs were unearthed.

### DIFF
--- a/model/advanced_controls.py
+++ b/model/advanced_controls.py
@@ -803,7 +803,7 @@ class AdvancedControls:
     def soln_lifetime_replacement_rounded(self):
         if self.soln_lifetime_capacity is not None:  # RRS
             # ROUND and decimal.quantize do not match Excel ROUND(), so we implemented one.
-            return excel_math.round(self.soln_lifetime_capacity / self.soln_avg_annual_use)
+            return excel_math.round_away(self.soln_lifetime_capacity / self.soln_avg_annual_use)
         elif self.soln_expected_lifetime is not None:  # LAND
             # LAND models input lifetime directly so I doubt we will come across rounding errors
             # i.e. expected_lifetime will probably be a whole number of years.
@@ -827,7 +827,7 @@ class AdvancedControls:
     def conv_lifetime_replacement_rounded(self):
         if self.conv_lifetime_capacity is not None:  # RRS
             # ROUND and decimal.quantize do not match Excel ROUND(), so we implemented one.
-            return excel_math.round(self.conv_lifetime_capacity / self.conv_avg_annual_use)
+            return excel_math.round_away(self.conv_lifetime_capacity / self.conv_avg_annual_use)
         elif self.conv_expected_lifetime is not None:  # LAND
             # LAND models input lifetime directly so I doubt we will come across rounding errors
             # i.e. expected_lifetime will probably be a whole number of years
--- a/model/advanced_controls.py
+++ b/model/advanced_controls.py
@@ -11,7 +11,6 @@ import typing
 
 import pandas as pd
 import pytest
-from warnings import warn
 from model import emissionsfactors as ef
 from model import excel_math
 from model.dd import REGIONS, MAIN_REGIONS
@@ -423,7 +422,7 @@ class AdvancedControls:
             "the SOLUTION.  In most cases this will be expressed as a cost per 'some unit of "
             "installation size' E.g., $10,000 per kw. In terms of transportation, this can be "
             "considered the total insurance, and maintenance cost per car.\n\n"
-            
+
             "Purchase costs can be amortized here or included as a first cost, but not both."),
         'excelref': 'SolarPVUtil "Advanced Controls"!I128; Silvopasture "Advanced Controls"!C92',
         })
@@ -469,7 +468,7 @@ class AdvancedControls:
     # npv_discount_rate: discount rate for Net Present Value calculations.
     #   SolarPVUtil "Advanced Controls"!B141
     npv_discount_rate: float = None
-  
+
     # emissions_use_co2eq: whether to use CO2-equivalent for ppm calculations.
     #   SolarPVUtil "Advanced Controls"!B189
     # emissions_grid_source: "IPCC Only" or "Meta Analysis" of multiple studies.
--- a/model/emissionsfactors.py
+++ b/model/emissionsfactors.py
@@ -122,7 +122,7 @@ class ElectricityGenOnGrid:
         elif self.ac.emissions_grid_range == GRID_RANGE.MEAN:
             result.loc[:, "World"] = grid.loc[:, "medium"].values
         else:
-            raise ValueError(f"Invalid ac.emissions_grid_range {ac.emissions_grid_range}")
+            raise ValueError(f"Invalid ac.emissions_grid_range {self.ac.emissions_grid_range}")
 
         # Generation mixes from the AMPERE/MESSAGE WG3 BAU scenario, direct and
         # indirect emission factors by fuel from the IPCC WG3 Annex III Table A.III.2
--- a/model/emissionsfactors.py
+++ b/model/emissionsfactors.py
@@ -7,7 +7,6 @@ and other factors relating to emissions and pollutants.
 from functools import lru_cache
 import enum
 import pandas as pd
-import model.dd as dd
 
 CO2EQ_SOURCE = enum.Enum('CO2EQ_SOURCE', 'AR5_WITH_FEEDBACK AR4 SAR')
 GRID_SOURCE = enum.Enum('GRID_SOURCE', 'META IPCC')
--- a/model/excel_math.py
+++ b/model/excel_math.py
@@ -1,20 +1,15 @@
 import math
 
 
-
-def round(x):
-    is_neg = False;
-
-    # Excel rounds away from zero, so 1.5 rounds to 2 and -1.5 rounds to -2.
-    if x < 0.0:
-        is_neg = True
-        x = math.fabs(x)
-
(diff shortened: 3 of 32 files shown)


avishayp:
For ref: current output of `python -m pylint --errors-only data model solution tests tools ui`
---

```
************* Module model.advanced_controls
model/advanced_controls.py:889:12: E1101: Instance of 'AdvancedControls' has no 'incorrect_cached_values' member
 (no-member)
************* Module model.customadoption
model/customadoption.py:88:19: E1133: Non-iterable value df.columns is used in an iterating context (not-an-iter
able)
************* Module model.tam
model/tam.py:213:22: E1133: Non-iterable value regional.columns is used in an iterating context (not-an-iterable
)
model/tam.py:422:22: E1133: Non-iterable value result.columns is used in an iterating context (not-an-iterable)
************* Module model.adoptiondata
model/adoptiondata.py:158:22: E1133: Non-iterable value regional.columns is used in an iterating context (not-an
-iterable)
model/adoptiondata.py:309:26: E1133: Non-iterable value df.columns is used in an iterating context (not-an-itera
ble)
model/adoptiondata.py:318:22: E1133: Non-iterable value df.columns is used in an iterating context (not-an-itera
ble)
************* Module model.helpertables
model/helpertables.py:120:19: E1133: Non-iterable value adoption.columns is used in an iterating context (not-an
-iterable)
************* Module model.metaclass_cache
model/metaclass_cache.py:42:14: E1120: No value for argument 'item' in unbound method call (no-value-for-paramet
er)
model/metaclass_cache.py:44:32: E1120: No value for argument 'item' in unbound method call (no-value-for-parameter)
model/metaclass_cache.py:46:32: E1120: No value for argument 'item' in unbound method call (no-value-for-parameter)
model/metaclass_cache.py:47:32: E1120: No value for argument 'item' in unbound method call (no-value-for-parameter)
************* Module tools.integration_test_xls_extract
tools/integration_test_xls_extract.py:13:0: E0401: Unable to import 'xlwings' (import-error)
tools/integration_test_xls_extract.py:23:4: E0401: Unable to import 'appscript.reference' (import-error)
ERROR: InvocationError for command /Users/avishay/Work/CC/solutions/.tox/common/bin/python -m pylint --errors-only data model solution tests tools ui (exited with code 2)
______________________________________ summary _______________________________________
ERROR:   lint: commands failed
```

`tox -e lint` uses an arbitrary cherry-picked list of W/E codes, which already provide false positives.

model/advanced_controls.py
```diff
@@ -804,7 +803,7 @@ def soln_lifetime_replacement(self):
     def soln_lifetime_replacement_rounded(self):
         if self.soln_lifetime_capacity is not None:  # RRS
             # ROUND and decimal.quantize do not match Excel ROUND(), so we implemented one.
-            return excel_math.round(self.soln_lifetime_capacity / self.soln_avg_annual_use)
+            return excel_math.round_away(self.soln_lifetime_capacity / self.soln_avg_annual_use)
```
DentonGentry: <strike>I don't think this works, https://github.com/ProjectDrawdown/solutions/blob/master/model/excel_math.py does not have a round_away method. We could add one, or rename the round() method in excel_match.py to round_away().

We have a model/excel_math.py implementation because Excel's behavior in rounding does not match Python's math.round(), leading to failures in various solutions run through tests/test_excel_integration.py.</strike>


Ok, saw the change in model/excel_math.py later.

model/emissionsfactors.py
```diff
@@ -123,7 +122,7 @@ def conv_ref_grid_CO2eq_per_KWh(self):
         elif self.ac.emissions_grid_range == GRID_RANGE.MEAN:
             result.loc[:, "World"] = grid.loc[:, "medium"].values
         else:
-            raise ValueError(f"Invalid ac.emissions_grid_range {ac.emissions_grid_range}")
+            raise ValueError(f"Invalid ac.emissions_grid_range {self.ac.emissions_grid_range}")
```
DentonGentry: I guess the ValueError case doesn't have a test.

model/tam.py
```diff
@@ -446,8 +444,8 @@ def pds_tam_per_region(self):
                 result[region] = self.forecast_trend(region_pds).loc[:, 'adoption']
                 lmh = self.forecast_low_med_high(region)
                 if result.dropna(axis=1, how='all').empty or lmh.dropna(axis=1, how='all').empty:
-                    result[region] = self.forecast_trend_global().loc[:, 'adoption']
-                    lmh = self.forecast_low_med_high_global()
+                    result[region] = self.forecast_trend(region='World', trend='Linear').loc[:, 'adoption']
+                    lmh = self.forecast_low_med_high(region='world')
```
DentonGentry: Ok, clearly another code path which needs a test.

model/excel_math.py
```diff
@@ -1,20 +1,15 @@
...
-    halfway = math.floor(x) + 0.5
-    x = math.floor(x) if x < halfway else math.ceil(x)
-    if is_neg:
-        x = -x
-    return int(x)
+def round_away(x):
+    """
+    excel rounds away from zero on breakeven
+    so 0.5 ==> 1.0, -0.5 ==> -1.0
+    python2 does the same, hence this code
```
DentonGentry: Interesting, I didn't know rounding behavior is different in python2 vs python3.

The tests in model/tests/test_excel_math.py covers the cases noticed in various solutions. The 2 hour test run with tests/test_excel_integration.py would check if this introduces any other subtle rounding issues impacting the solutions.

About 15 minutes after you sent this pull request, I checked in a change to the Github Actions workflow to run the test suite on pull requests and cleanly skip uploading codecov.io results if the secret key isn't available. So unfortunately it missed this one.

solution/airplanes/__init__.py
```diff
@@ -207,8 +207,6 @@ def __init__(self, scenario=None):
         high_sd_mult=1.0, low_sd_mult=1.0,
         total_adoption_limit=pds_tam_per_region)
 
-    ref_adoption_data_per_region = None
```
DentonGentry: I'd request to revert the changes in solution/*/\_\_init\_\_.py, as these are all code generated from the Excel files and will just be put back in the next time the code generator runs.

There is a possible combination of settings within a solution which will use ref_adoption_data_per_region, so the code generator always emits a None at the top and will use it later if those conditions hold.

solution/silvopasture/scenario_extract.py
```diff
@@ -5,15 +5,15 @@
 import sys
```
DentonGentry: I don't even remember what this file is. I suspect it is dead code.

solution/solarpvutil, solution/solarpvroof, and solution/silvopasture were the first three solutions worked on and were first implemented before we had the code generator. They have since been updated using the code generator to replace the original. I bet scenario_extract.py is a vestigial remnant of the first implementation.

solution/temperateforests/__init__.py
```diff
@@ -324,7 +324,8 @@ def __init__(self, scenario=None):
     commit_9_mha_new = max_land_wri - (commit_13_dec_2016 * commit_13_dec_2016_tmr)
     #         new commitments in temperate TMR + additional degraded land available from the Forest
     #         Protection solution, for which exact area has to be estimated.
-    final_adoption_9 = (commit_13_dec_2016_mha * intact_13_dec_2016) + (commit_9_mha_new * future_9_tmr)
+    # unused variable final_adoption_9
+    # final_adoption_9 = (commit_13_dec_2016_mha * intact_13_dec_2016) + (commit_9_mha_new * future_9_tmr)
```
DentonGentry: Line 338 (not present in the diff for code review) references final_adoption_8 but is likely supposed to be final_adoption_9.

This means that the researcher who defined the custom adoption data for the temperateforests solution didn't use them all in the active scenarios, so we didn't notice the breakage in tests/test_excel_integration.py. There are a few places like that, which define data that doesn't get used.

tests/test_excel_integration.py
```diff
@@ -6,12 +6,8 @@
 """
 
 import functools
-import os.path
```
DentonGentry: This test has gone through so many re-implementations I'm surprised there isn't more dead code in it.

tools/util.py
```diff
@@ -47,8 +45,12 @@ def to_filename(name):
     return re.sub(' +', '_', re.sub('[^a-zA-Z0-9' '\n]', ' ', name)).strip('_')
 
 
-def pretty_print_table(df):
-    """ Prints a nice-looking DataFrame """
-    from tabulate import tabulate
-    print(tabulate(df, headers='keys', tablefmt='psql', stralign='center', numalign='center', disable_numparse=False,
-                   floatfmt='.3f'))
+# method not used anywhere in code
```
DentonGentry: Recommend removing it altogether rather than commenting it out.

DentonGentry:
I think I'd be ok with merging this as is, if you think it is ready.
I'll plan to do a couple things afterwards:
+ I'll write test cases for the two code paths noted above which definitely should have them
+ I'll also remove solution/silvopasture/scenario_extract.py, as dead code
+ I'll fix solution/temperateforests/__init__.py, and uncomment final_adoption_9 again

avishayp:
Should be good to go.

Continuation:
 * Fix `tox -e lint` to ignore `solution/*/__init__.py`.
 * Add pylint to workflow with a carefully curated list of flags
 * Move to #48

DentonGentry:
https://github.com/ProjectDrawdown/solutions/pull/58 contains the items I was going to followup on.

PR #61, opened 2020-03-03
Force lint

avishayp:
### in this PR
 * Complete #47 (for now) by adding `pylint` to workflow
 * Fix #48 by autopepping codebase (excluding `__init__.py` files)

This should be a minimal-risk PR, bar the options that workflow-enforced linting will annoy future contributors.

PR #64, opened 2020-03-05
11 show warnings

### LINKED ISSUE #11: pytest to show DeprecationWarnings in our own code

avishayp:
### in this PR

 * Fix #11 
 * Try reducing noise by ignoring `ipy*` modules warnings
 * Clean some obvious warnings from code

The short suite give me 23 warnings now and is readable. Waiting to see the full suite output.

### DIFF
--- a/.github/workflows/drawdown.yml
+++ b/.github/workflows/drawdown.yml
@@ -28,6 +28,9 @@ jobs:
       run: |
         pip install tox
         tox -e ci
+    - name: pylint
+      run: |
+        tox -e lint
     - name: Export to codecov.io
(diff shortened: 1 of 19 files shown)


avishayp:
This should be merged after #61

PR #77, opened 2020-03-19
[Issue #6] Run cronjob health surveys

4dahalibut:
I accidentally refactored a bit too. LMK how this goes.


tools/health/landsurvey.py
```diff
@@ -1,23 +1,27 @@
...
 
-import solution.factory
 import model.aez
 import model.dd
+import solution.factory
 import tools.util
 
 pd.set_option('display.expand_frame_repr', False)
-datadir = pathlib.Path(__file__).parents[2].joinpath('data')
+datadir = pathlib.Path('data')
```
DentonGentry: I believe it previously used Path(\_\_file\_\_).parents[2] so that it could be run from any directory and find its data file. I think doing it this way, though simpler, only works if $PWD is tools/health. Though the cronjob would do so, it would be nice to still be able to easily run it manually if (for example) we're checking in new versions of solution models.

tools/health/survey.py
```diff
@@ -38,11 +40,11 @@ def scenario_survey(outfile):
...
+    return results
+
 
 if __name__ == '__main__':
-    (_, outfile) = tempfile.mkstemp(prefix='survey_', suffix='.csv',
-                                    dir=os.path.join('data', 'health'))
-    scenario_survey(outfile)
+    # Write the results of the survey out
+    scenario_survey().to_csv(path_or_buf=FILENAME_TO_WRITE_TO)
     # print the file name on exit, allows the calling script to decide whether to check it in.
```
DentonGentry: Can delete this comment and the print() on the next line. Since it is writing the output in place, it doesn't need to do this any more.

tools/health/survey.py
```diff
@@ -1,13 +1,15 @@
...
-import os.path
-import tempfile
 import numpy as np
 import pandas as pd
 import scipy.stats
+
 import solution.factory
 
-def scenario_survey(outfile):
+FILENAME_TO_WRITE_TO = 'data/health/survey.csv'
```
DentonGentry: Recommend using Pathlib for this as I don't think a slash-delimited path works on Windows systems (which some developers have used).

data/health/landsurvey.csv
```diff
@@ -14,8 +14,8 @@ peatlands,99.03852328101632,2.957079080830361,-0.0,core,False,False,False,0,0,0,
 perennialbioenergy,100.0,0.5965099622702997,25.785808851943987,core,False,False,False,1,0,0,0
 regenerativeagriculture,71.31869674706621,6.317660007446539,-13.42572289410737,yield,True,False,False,0,0,0,0
 riceintensification,100.0,0.44619757678573857,-0.21293679151627978,yield,False,False,False,0,0,0,0
-silvopasture,59.078159324925544,7.488335317552567,4.835255217612302,yield,True,False,False,2,0,0,0
-temperateforests,74.10377705777432,1.5714226231531045,-0.0,core,False,False,False,0,0,0,0
+silvopasture,59.078159324925544,7.488335317552567,4.117918039226432,yield,True,False,False,2,0,0,0
```
DentonGentry: To be honest I don't understand why the values are changing here as I thought we'd run the landsurvey manually... which I suppose shows the value of automating it. Well done!

.github/workflows/survey-health.yml
```diff
@@ -0,0 +1,42 @@
...
+    
+        # Optional glob pattern of files which should be added to the commit
+        file_pattern: data/health/*
+    
+        # Optional local file path to the repository
+        repository: .
+    
+        # Optional commit user and author settings
+        commit_user_name: Health Survey Runner Bot
+        commit_user_email: bot@computer.edu
```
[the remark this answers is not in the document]
4dahalibut: Sure!

tools/health/landsurvey.py
4dahalibut: Sure I'll change this back.

tools/health/survey.py
4dahalibut: Okay

tools/health/survey.py
4dahalibut: Makes sense

data/health/landsurvey.csv
4dahalibut: The values seemed to be changing non-deterministically...

DentonGentry: Ok, that is interesting. I think we just proceed, and if the numbers keep changing we'll look into why.

Issue #5, opened 2020-02-08
Selenium tests for Jupyter


DentonGentry:
Very little of the code used in the Jupyter UI is embedded within the notebook file itself. Most of the code is in separate Python files, with some very basic unit tests. Nonetheless, test coverage is not great.

We would benefit greatly from automated UI tests like Selenium, or a similar framework. Setting up such tests would not require extensive familiarity with the Drawdown codebase or models, it would mostly require familiarity with construction of such tests.

Even if we deprecate the Jupyter notebook UI, we'd bring along the test framework and set up tests for the new UI.

jonathan-s:
Some notes on testing jupyter notebooks -> https://blog.thedataincubator.com/2016/06/testing-jupyter-notebooks/ 

That is to say using selenium may not be the best approach. Unless it's purely UI that is being tested (I still haven't read much of the project, so I may be completely wrong here, but at first glance it sounds like pure unittest would be better than using selenium).

DentonGentry:
nbconvert does work to render the Overview page of the notebook, both the Voilà  version and the Jupyterlab version:

`jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=60 --output rendered ./Drawdown.ipynb`

`jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=60 --output rendered ./Drawdown.ipynb`

This is what the rendered Overview page would look like.

<img width="749" alt="Screen Shot 2020-02-26 at 7 22 03 PM" src="https://user-images.githubusercontent.com/1140756/75409348-540e8880-58cd-11ea-8f05-6954f0d1273c.png">

Clicking the checkbox triggers the rest of the tabs to be run: https://youtu.be/RAQSnhm_tS4

I think that testing the functionality past the Overview page would require a UI testing framework rather than nbconvert alone.

4dahalibut:
I'll check this one out.

DentonGentry:
https://github.com/ProjectDrawdown/solutions/pull/80
https://github.com/ProjectDrawdown/solutions/pull/86
https://github.com/ProjectDrawdown/solutions/pull/87

Calling this one done.

PR #80, opened 2020-03-22
[#5] First Smoketest of Voila App

4dahalibut:
https://github.com/ProjectDrawdown/solutions/issues/5 
Here lies the simplest possible smoketest configuration I could think of. 

Notice that I added a geckodriver binary to the repo. Sketchy. 

Also I may have to install firefox into the github workflow for this to work. 

Also you might want me to also use chrome but I feel like this crowd should just be using firefox anyways right. 

Also I'm introducing the start.sh and stop.sh paradigms here, is this something you're down with? It's what we do at my job.


tests/test_smoke.py
```diff
@@ -0,0 +1,23 @@
...
+    options = Options()
+    options.headless = True
+    driver = webdriver.Firefox(executable_path='bin/geckodriver', options=options)
+    yield driver
+    driver.close()
+
+
+def test_voila_started(driver):
+    driver.get("http://localhost:8866/")
+    time.sleep(10)
```
4dahalibut: Sketchy

tests/test_smoke.py
```diff
@@ -0,0 +1,23 @@
...
+    assert "Drawdown" in driver.title
+    elem = driver.find_element_by_xpath('//div[contains(text(), "Solution")]')
```
4dahalibut: Makes sure that the first page of solution menus actually loads at all.

stop.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+pkill voila
```
4dahalibut: sketchy

.github/workflows/drawdown.yml
```diff
@@ -31,6 +31,9 @@ jobs:
     - name: pylint
       run: |
         tox -e lint
+    - name: Jupyter Smoketest
```
4dahalibut: Runs sequentially, but the test doesn't take that much overhead.

start.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+./stop.sh
+voila --enable_nbextensions=True ./VoilaDrawdown.ipynb --no-browser &
```
4dahalibut: it was complaining that the whitelist thing listed in the readme wasn't being parsed.

DentonGentry: Voila has updated versions several times since that was written, might no longer be correct. I'll look into that separately.

stop.sh
DentonGentry: For running within the container for the test it would be fine, the entire container will exit at the end anyway.

To run by developers this would probably need to be more targeted to just the process started by start.sh. I think that doesn't need to be done in this first PR.

start.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
DentonGentry: Could we put start.sh and stop.sh in tests/ui (new directory)?

tests/test_smoke.py
DentonGentry: It seems like this could be `driver.implicitly_wait(10)` to put a 10 second timeout on each driver access which follows, to give the page some time to render.

As we know that the Markdown cell at the top of the page appears almost immediately but the controls which are rendered from code in ui/*.py take longer, maybe `driver.implicitly_wait(30)`

tox.ini
```diff
@@ -20,7 +20,12 @@ commands =
 # runs on github workflow
 [testenv:ci]
 commands =
-    python -m pytest --cov=. --cov-report term --cov-report=xml
+    python -m pytest --cov=. --cov-report term --cov-report=xml --ignore=tests/test_smoke.py
+
+[testenv:smoke]
+commands =
+    ./start.sh
+    python -m pytest tests/test_smoke.py
```
DentonGentry: Recommend running stop.sh. If running as a github workflow it would be killed when the container is stopped, but a developer running it on their workstation would accumulate browser processes.

I imagine we'll work on stop.sh a bit before other developers might routinely begin running Selenium.

.github/workflows/drawdown.yml
DentonGentry: When we're sufficiently confident, I'd plan to move this into the CI Tox so that code coverage statistics will include the browser test. The current ui/tests/test_*.py does little more than verify that the code doesn't die, over time we could ensure that the UI code does what is expected.

It is good to add this is as separate item right now, though, and let it run for a while.

requirements.txt
```diff
@@ -14,6 +14,7 @@ pytest>=5.0.0
 pytest-cov>=2.7.1
```
DentonGentry: Is there a way to add geckodriver as a requirements.txt dependency? I have to say it makes me uncomfortable to have a binary checked in, not so much for github CI but for developers who might run it locally:
- developers working on this have used Windows and MacOS, the Linux binary wouldn't work for them.
- developers running Linux have no way to verify that this is what it says it is.

https://stackoverflow.com/questions/56565278/could-not-find-a-version-that-satisfies-the-requirement-geckodriver-0-24-0-fro implies that it is possible to get geckodriver from requirements.txt, so at least the level of trust required moves from a binary checked into this repository to trusting pip as a source of packages.

tests/test_smoke.py
DentonGentry: This is excellent. We'd build on it over time.

start.sh
4dahalibut: I think start.sh and stop.sh can be user intuitive ways to control the application, so we can change the readme to just say `use start.sh` instead of running the crazy voila command.

requirements.txt
4dahalibut: This might be an option https://github.com/juancroldan/autoselenium seems very unused.

4dahalibut: This one seems like the most used. geckodriver isn't in pypi (anymore) https://github.com/rasjani/webdrivermanager

Issue #2, opened 2020-02-08
Make vma.py support XLS files


DentonGentry:
[model/vma.py](https://github.com/ProjectDrawdown/solutions/blob/master/model/vma.py) implements Variable Meta-Analysis, where we produce a value for an input variable like the yield for soybeans or the cost for a megawatt from natural gas powerplants. Researchers at Project Drawdown vet data sources as inputs, ideally multiple sources, and vma.py collects these sources and produces a single resulting value. This is typically the mean, though +/- multiples of the standard deviation is also common.

Right now in the \_\_init\_\_ method we call VMA._read_csv, which reads in a CSV file. The existing CSV files were produced via the [vma_xls_extract.py](https://github.com/ProjectDrawdown/solutions/blob/master/tools/vma_xls_extract.py) code generator from the original Drawdown Excel models.

This issue concerns adding support for XLS files for VMA input, to allow researchers to more easily perform data normalizations like currency or unit conversion.

The desired steps are:

1. In vma.py VMA::\_\_init\_\_, check the file extension of the data source and add support for *.xlsx and *.xlsm

1. For CSV we require a separate file for each VMA. For Excel we want to support multiple VMA definitions within one file, to allow the researcher to implement their needed conversions once not have copies in multiple files.  
Therefore, the code should open the Excel file and then search for the definition of its VMA. Searching for the name of the VMA within the first sheet of the workbook, and figuring out where the VMA definition is below that, is preferred.

1. advanced_controls.py, which instantiates VMA objects, knows the human-readable Title of the VMA it is looking for. The VMA::\_\_init\_\_ does not currently receive the Title as a parameter, but it can be added.  
Please do not add a default value, the codebase is small enough that we can update all existing callers to pass in a proper value. If the backing file is CSV, the title argument may just not be used.

1. Add unit tests to model/tests/test_vma.py. Add an Excel file for use in the test in model/tests/data.

1. Please use Pandas read_excel() and ensure it works with the 'xlrd' backend, as we already have dependencies on xlrd in the tree. We do not currently have any dependencies on other Excel+python packages like openpyxl, and would prefer not to add new dependencies without a really good reason.

DentonGentry:
It would be great if a single Excel file for a solution could supply VMA (issue #2), customadoption/adoptiondata/etc (issue #3), plus allow additional sheets for various and sundry stuff specific to the solution which doesn't benefit from being done within Python.

[...]

PR #126, opened 2020-04-24
Adding xlsx/xlsm support to vma.VMA

FranzEricSchneider:
Intended to address https://github.com/ProjectDrawdown/solutions/issues/2

`make test` passed:
```
(solutions) solutions $ make tests
...
======================== 433 passed, 156 deselected, 260 warnings in 123.03s (0:02:03) ========================
  py38: commands succeeded
  congratulations :)
```

`make alltests` passed:
```
(solutions) solutions $ make alltests
...
======================== 589 passed, 455 warnings in 5034.73s (1:23:54) ========================
  all: commands succeeded
  congratulations :)
```

In `vma.VMA.__init__`, checks the filetype and supports `*.xlsx` and `*.xlsm`. Reads the excel file into a workbook with `xlrd.open_workbook`, then uses existing `VMAReader` class (with a split up function) to get the specific title.

Adds unit tests to `model/tests/test_vma.py` and `tools/tests/test_vma_xls_extract.py`, along with `xlsx`/`xlsm` files in `model/tests/data`.


model/vma.py
```diff
@@ -66,37 +94,132 @@ def __init__(self, filename, low_sd=1.0, high_sd=1.0, discard_multiplier=3,
...
+
+        # Pull the desired table from this workbook
+        try:
+            (xl_df, use_weight, summary) = \
+                vma_reader.xls_df_dict(alt_vma=alt_vma, title=title)[title]
+        except KeyError:
+            # The title wasn't available in the given workbook. Read all titles
+            # to give the user a hint.
+            full_dict = vma_reader.xls_df_dict(alt_vma=alt_vma)
+            raise ValueError(
```
FranzEricSchneider, line 157: I wanted to include the suggested titles in the error message to help the user figure out what they were trying to get. Given the complicated nature of the titles, it's easy to have one word mispelled, or an extra space, or a small change between Excel text and Python text. This could raise a simpler `ValueError` if desired. Here's what it looks like:
```
E   ValueError: Title 'Not a real title' not available in /home/eric/projects/solutions/model/tests/data/silvopasture.xlsx.
E   Options:
E       Current Adoption
E       CONVENTIONAL First Cost per Implementation Unit
E       SOLUTION First Cost per Implementation Unit
E       CONVENTIONAL Operating Cost per Functional Unit per Annum
E       SOLUTION Operating Cost per Functional Unit per Annum
E       CONVENTIONAL Net Profit Margin per Functional Unit per Annum
E       SOLUTION Net Profit Margin per Functional Unit per Annum
E       Yield from CONVENTIONAL Practice
E       Yield Gain (% Increase from CONVENTIONAL to SOLUTION)
E       Electricty Consumed per CONVENTIONAL Functional Unit
E       SOLUTION Energy Efficiency Factor
E       Total Energy Used per SOLUTION functional unit
E       Fuel Consumed per CONVENTIONAL Functional Unit
E       Fuel Reduction Factor SOLUTION
E       t CO2-eq (Aggregate emissions) Reduced per Land Unit
E       t CO2 Reduced per Land Unit
E       t N2O-CO2-eq Reduced per Land Unit
E       t CH4-CO2-eq Reduced per Land Unit
E       Indirect CO2 Emissions per CONVENTIONAL Implementation OR functional Unit -- CHOOSE ONLY ONE
E       Indirect CO2 Emissions per SOLUTION Implementation Unit
E       Sequestration Rates
E       Sequestered Carbon NOT Emitted after Cyclical Harvesting/Clearing
E       Disturbance Rate
E       Percent silvopasture area to the total grassland area (including potential)
```

model/vma.py
```diff
@@ -66,37 +94,132 @@ def __init__(self, filename, low_sd=1.0, high_sd=1.0, discard_multiplier=3,
...
+            readable_df: dataframe (from CSV or Excel file) using the standard
+                long-form column names
+
+        Populates self.source_data with readable_df directly. Populates self.df
+        with a series of renamed columns, along with a few data cleanup steps.
+        """
+        self._validate_readable_df(readable_df)
+        if readable_df is None:
+            if self.title is None:
+                source = "\n\tFile: {}\n".format(self.filename)
```
DentonGentry, line 185: Could I request the use of f-strings? We use them all over (which means we already require Python 3.6+).

```source = f"\n\tFile: {self.filename}\n"```

If there were any existing uses of .format() in this file, that wasn't deliberately retained we just started with Python 3.5 way back.

File: tools/vma_xls_extract.py [hunk 3]
```diff
@@ -127,52 +148,92 @@ def read_xls(self, csv_path=None, alt_vma=False):
...
-        vma_df = pd.DataFrame(columns=['Filename', 'Title on xls', 'Has data?', 'Use weight?'],
-                index=idx)
+        return df_dict
+
+    def read_xls(self, csv_path=None, alt_vma=False):
+        """
+        Reads the whole Variable Meta-analysis xls sheet. If a CSV path is
+        given, writes out a series of CSVs, of the form <table title>.csv and
+        VMA_info.csv (contains summary data for each VMA).
+        Note this currently only works for LAND solutions.
```
DentonGentry, line 183: I know this comment is in the original file but it isn't accurate any more, the VMA extraction works on all of the solutions not just land. Could you delete line 183?

model/vma.py
```diff
@@ -13,22 +19,35 @@ def populate_fixed_summaries(vma_dict, filename):
...
         fixed_low = row.get('Fixed Low', np.nan)
-        fixed_summary = None
-        if not pd.isna(fixed_mean) and not pd.isna(fixed_high) and not pd.isna(fixed_low):
-            fixed_summary = (fixed_mean, fixed_high, fixed_low)
+        fixed_summary = check_fixed_summary(fixed_mean, fixed_high, fixed_low)
         if fixed_summary is not None:
             vma_dict[title].fixed_summary = fixed_summary
 
 
+def check_fixed_summary(*values):
```
DentonGentry, line 38: Could I ask if this caused problems?

I don't know if there were fixed summaries with NaN in them; since this PR passed tests/test_excel_integration, I guess there aren't any. However it is very common for the Excel VMAs to be #DIV/0!, if the VMA is not needed in the given solution. The human researcher has no cause to notice or do anything about it.

I think we can proceed with check_fixed_summary() as it is here, and if we find a reason to relax the assert and allow NaN we can do so later.

model/vma.py
```diff
@@ -66,37 +94,132 @@ def __init__(self, filename, low_sd=1.0, high_sd=1.0, discard_multiplier=3,
...
+
+            # Instantiate VMA with various file types
+            if isinstance(filename, io.StringIO):
+                self._read_csv(filename=filename)
+            elif filename.suffix == '.xlsx' or filename.suffix == '.xlsm':
+                self._read_xls(filename=filename, title=title)
+            else:
+                # Fall back on this as the default behavior, there are a
+                # variety of use cases that have been built-in, they do not
+                # all end with '.csv'
```
DentonGentry, line 111: Oops: that wasn't intentional, though I can certainly believe it happened. Later I may take this default back out and see what breaks, and try to ensure they all have file extensions.

model/vma.py
FranzEricSchneider: Sure

tools/vma_xls_extract.py
FranzEricSchneider: Sounds good, I wondered about that

model/vma.py
FranzEricSchneider: Actually, now that you mention it, there are only a few spots. I originally saw it all over the place and didn't think of fixing it, but the `isinstance(filename, io.StringIO)` statement actually caught nearly all of them. These are the only remaining offenders, from `test_vma.py`.
```
E               ValueError: /tmp/tmp1ncvbfsf not a CSV
E               ValueError: /tmp/tmp2jgeuxxr not a CSV
```
Looks like I can clean those up by adding `suffix='.csv'` to the `tempfile.NamedTemporaryFile` calls.

model/vma.py
FranzEricSchneider: You know, the only time the `return None` case comes up in `make tests` is when it is specifically checked in a test. I just made the function because this is what the previous behavior was and I wanted to move it around but preserve it.

Want me to either take it out or try it again with `make alltests` to see if it shows up there?

model/vma.py
```diff
@@ -66,37 +94,132 @@ def __init__(self, filename, low_sd=1.0, high_sd=1.0, discard_multiplier=3,
...
+        Meta-analysis sheet) to instantiate this VMA.
+
+        Arguments:
+            filename: pathlib.Path to an Excel file
+            title: string matching VMA name in the Variable Meta-analysis sheet
+
+        Populates self.source_data, self.df, and self.fixed_summary if the
+        required values are present.
+        """
+        workbook = xlrd.open_workbook(filename=filename)
```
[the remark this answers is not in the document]
FranzEricSchneider: I sort of tried that early on, in the first version of what I was doing I
1. read the workbook
2. populated all VMA tables as df instances
3. used only the title I wanted as the `source_data` for the current VMA and threw away the rest.

Later I changed step 2 to only read the table we wanted for time reasons, populating all tables took about as long as reading the workbook once.

Part of my problem is I don't know how/where `vma.VMA('xlsx')` is going to be called. If there's some orchestrating code that is going to read a bunch of VMAs in, it probably makes sense for that code to open the workbook and then pass that into each VMA instance being created. There could be a `wb=None` argument to `VMA.__init__` that is used only by `_read_xls`, and the class popuulates its own workbook if that is `None`

model/vma.py
DentonGentry: I'd say not to do that just yet: if we can reasonably require that VMAs have a proper value, it would be nice to not have to deal with NaNs. If we find that this doesn't work, we can relax it later.

model/vma.py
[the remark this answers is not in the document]
DentonGentry: Each solution, like Solar Farms or Regenerative Agriculture, has a directory in solutions where the \_\_init\_\_.py file instantiates a list of VMA objects. The \_\_init\_\_.py file is code generated by tools/soluton_xls_extract.py from the solution's Excel file, and currently uses tools/vma_xls_extract.py to save each VMA's data to solution/name/vma_data/\*.csv. \_\_init\_\_.py instantiates those CSV filenames:

```
 VMAs = {
   'Current Adoption': vma.VMA(
       filename=THISDIR.joinpath("vma_data", "Current_Adoption.csv"),
       use_weight=False),
   'CONVENTIONAL First Cost per Implementation Unit': vma.VMA(
       filename=THISDIR.joinpath("vma_data", "CONVENTIONAL_First_Cost_per_Implementation_Unit.csv"),
       use_weight=True),
   'SOLUTION First Cost per Implementation Unit': vma.VMA(
       filename=THISDIR.joinpath("vma_data", "SOLUTION_First_Cost_per_Implementation_Unit.csv"),
       use_weight=True),
```

One more detail is that there are a few VMAs which are the same across multiple solutions, for example all of the energy solutions have some VMAs about the overall demand for electricity which are the same in all energy solutions. I identified those manually, and checked in the CSV files in data/\*/vma\_\*.csv. The code generator is smart enough to check if the CSV file already exists in data/\* and will use it instead of its own solution/name/vma_data directory.

Going forward, I'd say:
+ we'd still instantiate a list of VMAs in the \_\_init\_\_.py for each solution.
+ the VMA.vma() would specify an XLS file if one exists, or CSV if that exists.
+ we may want to cache the contents of XLS files in model/vma.py (or somewhere) to reduce the overhead, but that can wait for a subsequent effort not block this one.

So for this topic, I'd say it would be best to merge this PR so we don't create more merge conflicts — as I see you already had to resolve once — and consider how to approach caching the XLS content in a future change.

I'll plan to merge this PR at the end of the (California) afternoon today unless I hear otherwise.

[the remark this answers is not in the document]
FranzEricSchneider: Sounds good, thanks. Are there any other open issues somewhat related to this that you'd be excited about? I'd be interested in picking up another issue. Doesn't have to be related to this, this just happens to be where I have a little exposure now :)

PR #137, opened 2020-05-05
Add model health dashboard

klemag:
Hi @DentonGentry - that's still a WIP but wanted to know what you think before going too far with this.

I've added code to generate a static page using jinja2 and saving as index.html. Github pages picks up index.html (from master branch) and routes it to http://username.github.io/repository as explained [here](https://pages.github.com/).

I've made a test repo to check how it looks like: https://klemag.github.io/solutions_test/

Obviously not the nicest dashboard so far, but we've got all the components to add more charts / customise it as we want.

If you're happy with this solution, I just need to add the remaining plots that were in the notebook, and add the command to regenerate index.html to a cron (and maybe make the dashboard a bit nicer)

Let me know what you think


index.html
```diff
@@ -0,0 +1,158 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <title>Model Health Dashboard</title>
+</head>
+
+
+<script type="text/javascript" src="https://cdn.bokeh.org/bokeh/release/bokeh-2.0.2.min.js" integrity="sha384-ufR9RFnRs6lniiaFvtJziE0YeidtAgBRH6ux2oUItHw5WTvE1zuk9uzhUU/FJXDp" crossorigin="anonymous"></script>
```
DentonGentry, line 9: Bokeh is reasonable, sure. I had looked into a number of charting packages for Jupyter (https://codingrelic.geekhold.com/2019/02/line-graphs-in-jupyterlab.html) where Bokeh was one of the options. As we're moving this outside of Jupyter now, Bokeh has the advantage of not being strongly tied to Jupyter.

model_health/__main__.py
```diff
@@ -0,0 +1,12 @@
+from os.path import dirname, abspath
+import sys
+
+sys.path.append(dirname(dirname(abspath(__file__))))
```
DentonGentry, line 4: pathlib has been nice for things like this:
`mydir = pathlib.Path(__file__).parents[1]`

DentonGentry:
As an overall approach, this is great. Use of Jinja2 for templating and bokeh for graphs are both fine.

I concur with the approach of statically generating the files from a cronjob. There is an existing job in .github/workflows/survey-health.yml which regenerates the health CSV files this code is reading in. It might be reasonable to also add index.html generation to that job.

I do have one request: could the directory be named something other than "model_health", like just "health" or "dashboard"? I ask because otherwise my shell autocomplete for "model/" will not work anymore, and muscle memory is hard to change.

klemag:
@DentonGentry  it should be ready now. A few things:

- In the notebook comments were used to group together some charts. I used tabs on the dashboard, reusing roughly the same names, but not sure they are the most representative, I can update to whatever you think is best
- They cron should work, but I haven't been able to test it
- To see the dashboard on GitHub pages someone with admin rights needs to enable pages on master branch: https://pages.github.com/

dashboard/charts.py
```diff
@@ -0,0 +1,114 @@
...
+from bokeh.transform import cumsum
+import numpy as np
+import pandas as pd
+
+
+def make_pie_chart(data, cat_column, val_column, title, as_html=True):
+    data = data.copy()
+    data["angle"] = data[val_column] / data[val_column].sum() * 2 * pi
+    if data.shape[0] == 2:
+        data["color"] = Colorblind[3][:2]
```
DentonGentry, line 15: The pie chart of "Excel & Python" versus "Excel Only" used a yellow+blue color scheme because it looked like Pac-Man eating the solutions. I'm amused that the color scheme using bokeh.palettes.Colorblind happens to be sufficiently close to yellow + blue to look like Pac-Man.

dashboard/templates/index.html
```diff
@@ -0,0 +1,134 @@
...
+{% for chart_group in [summary, regional, land_survey, scenarios] %}
+{% for name, chart in chart_group.items() %}
+{{chart.script|safe}}
+{% endfor %}
+{% endfor %}
+
+<body>
+
+    <div class="dashboard container">
+        <img src="https://raw.githubusercontent.com/ProjectDrawdown/solutions/master/data/images/ProjectDrawdown.png"
```
DentonGentry, line 21: We're likely to forget that this dashboard references that image, and change or remove it in the future without noticing. If possible, recommend making a copy of it in dashboard/ somewhere to reference here.

index.html
```diff
@@ -0,0 +1,731 @@
...
+          else document.addEventListener("DOMContentLoaded", fn);
+        })();
+</script>
+
+
+
+<body>
+
+    <div class="dashboard container">
+        <img src="https://raw.githubusercontent.com/ProjectDrawdown/solutions/master/data/images/ProjectDrawdown.png"
```
DentonGentry, line 603: Similarly here, recommend using an image from dashboard/*

dashboard/charts.py
klemag: I didn't even notice that, I like it :D

dashboard/templates/index.html
klemag: Makes sense - I've changed it.

PR #173, opened 2020-06-07
Convert NE admin countries shapefile to Drawdown regions

olehb:
This script takes Natural Earth's `ne_10m_admin_0_countries.zip`, maps individual countries to Drawdown regions according to [admin_names.py](https://github.com/ProjectDrawdown/spatial-aez/blob/master/admin_names.py), and outputs GeoJSON, which only contains Drawdown regions. Information of individual countries is removed.

This PR is work-in-progress. Next step is to use [topojson-server](https://github.com/topojson/topojson-server) and [topojson-simplify](https://github.com/topojson/topojson-simplify) libraries to convert GeoJSON to TopoJSON or reasonable size (under 1Mb)

I found Python library for TopoJSON lacking in functionality and performance, so decided to use NodeJS implementation instead.

Opening this PR to validate this approach and clarify a few things regarding country-to-region mapping.


tools/ne_countries_to_drawdown_regions_geojson.py
```diff
@@ -0,0 +1,82 @@
...
+    dd_name = admin_names.lookup(ne_name)
+
+    if dd_name in SPECIAL_COUNTRIES:
+        return dd_name
+    else:
+        dd_region = admin_names.region_mapping.get(dd_name)
+        if dd_region is None:
+            return None
+        elif isinstance(dd_region, list):
+            # TODO: What to do if length of the list is more than 1?
```
olehb, line 42: Sometimes [region_mapping](https://github.com/ProjectDrawdown/spatial-aez/blob/master/admin_names.py#L89) dictionary maps a country to more than one region. For example, `"Czech Republic": ["Eastern Europe", "EU"]`. What to do in this case?

olehb:
TopoJSON file is produced like this:
```
> ./node_modules/topojson-server/bin/geo2topo ../solutions/ne_10m_admin_0_countries.json > ne_10m_admin_0_countries_topo.json
> ./node_modules/topojson-simplify/bin/toposimplify -P 0.035 ne_10m_admin_0_countries_topo.json > ne_10m_admin_0_countries_topo_simplified.json
```
Note, that simplification parameter `P` is [configurable](https://github.com/topojson/topojson-simplify#toposimplify_planar_quantile). Making it smaller makes output file smaller, but also removes fine details. Example output: [ne_10m_admin_0_countries_topo_simplified.json](https://github.com/olehb/solutions/blob/data/ne_admin_countries_to_drawdown_regions/data/ne_10m_admin_0_countries_topo_simplified.json)

tools/ne_countries_to_drawdown_regions_geojson.py
DentonGentry: We'd want the final TopoJSON file to have topologies for both Eastern Europe and for the EU. So it would need to add the Czech Republic to both of those regions, and once all countries have been processed it would need to remove the internal topology for Eastern Europe and separately remove the internal topology for the EU.

China, India, EU, and USA are considered special regions: the countries involved are in one of the main regions like OECD90, but also broken out separately because their emissions are so large.

DentonGentry:
Regarding the amount of topology simplification: this is only used for data visualizations in the user interface, and typically rendered at about 300 to 400 pixels square. We can set the simplification setting quite high, and toss out a lot of the fine detail, because no-one would be able to see those details in a small rendering.

If we ever have processing which would need to accurately account land areas we'd go back to the spatial-aez repository and implement the processing using the GeoTIFF files.

DentonGentry:
I have not merged this because it has "WIP" in the description, but it is good work and even in its current state would be the basis for further work on this. I'd like you to get author credit for the work done. Should I merge it as is?

If I don't hear back I'll merge it in a few days.

olehb:
- Resolved issue with same country being in multiple Drawdown regions.
- Added result file (I hope it's not too big, but I can remove it or make it smaller if needed)
- Added console output message about how to convert GeoJSON to simplified TopoJSON 

This PR is no longer WIP and is ready for review

tools/ne_countries_to_drawdown_regions_geojson.py
olehb: I resolved this issue by adding same country geometry into multiple regions

olehb:
If final datafile is not needed, I can close this one and open a new PR without datafile and with cleaner commit history from this branch: https://github.com/olehb/solutions/tree/feature/ne_admin_countries_to_drawdown_regions_clean

olehb:
Bumping this up... @DentonGentry

DentonGentry:
Yes, we'll merge this and use it for the next steps in the geo illustrations. Thank you!

PR #222, opened 2021-03-20
Update README with a hacky solution for database initialization

jhslinkman:


Sunishchal (approved):

jhslinkman:
Updated with a simpler approach, using `docker-compose exec db /bin/bash` to connect to the db container

PR #245, opened 2021-06-15
readded utility files

sutjin:
Readded some utility files that got deleted to allow for local development


brodavi (approved):
Ah yes, thanks @sutjin !  We might have been a bit too zealous in cleaning up :) :+1:

Issue #20, opened 2020-02-08
Rationalize regional support


DentonGentry:
A goal for Project Drawdown going forward is to support modelling efforts at scales more focussed than the entire world. It is a goal to support solution models where, for example, the top level is the United States, or India, or similar and the major regions are the states and territories.

There are two places in the Python model implementation which know the set of regions supported:

[model/dd.py](https://github.com/ProjectDrawdown/solutions/blob/master/model/dd.py):
```
REGIONS = [
    'World',
    'OECD90', 'Eastern Europe', 'Asia (Sans Japan)', 'Middle East and Africa',
    'Latin America',
    'China', 'India', 'EU', 'USA'  # special countries
]
MAIN_REGIONS = REGIONS[1:6]
SPECIAL_COUNTRIES = REGIONS[6:]
COUNTRY_REGION_MAP = {'China': 'Asia (Sans Japan)', 'India': 'Asia (Sans Japan)',
    'EU': 'OECD90', 'USA': 'OECD90'}
```

[model/emissionsfactors.py](https://github.com/ProjectDrawdown/solutions/blob/master/model/emissionsfactors.py):conv_ref_grid_CO2eq_per_KWh()
```
# Generation mixes from the AMPERE/MESSAGE WG3 BAU scenario, direct and
# indirect emission factors by fuel from the IPCC WG3 Annex III Table A.III.2
# https://www.ipcc.ch/pdf/assessment-report/ar5/wg3/ipcc_wg3_ar5_annex-iii.pdf
result.loc[:, "OECD90"] = 0.454068989
result.loc[:, "Eastern Europe"] = 0.724747956
result.loc[:, "Asia (Sans Japan)"] = 0.457658947
result.loc[:, "Middle East and Africa"] = 0.282243907
result.loc[:, "Latin America"] = 0.564394712
result.loc[:, "China"] = 0.535962403
result.loc[:, "India"] = 0.787832379
result.loc[:, "EU"] = 0.360629290
result.loc[:, "USA"] = 0.665071666
```

Implementing a different set of regions currently requires forking the codebase, or developing on a branch. There should instead be a way to accommodate multiple datasets and sets of regions in a single branch.

PR #247, opened 2021-06-15
Workbook Detail cards display accurate "last updated" data #20

brodavi:
adding an update to the workbook's `updated_at` when a user patches a
variation

### DIFF
--- a/api/routers/routes/workbook.py
+++ b/api/routers/routes/workbook.py
@@ -8,6 +8,7 @@ import uuid
 import aioredis
 import fastapi_plugins
 from typing import List, Optional, Any, Literal, Union
+from datetime import datetime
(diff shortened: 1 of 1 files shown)


brodavi:
just adding updated_at update to the api workbook when user updates a variation. UI relies on updated_at on the workbook to reflect updated info for workbook. but updating a variation doesn't actually update the workbook itself, so updated info was wrong in ui.

denised (commented):
I'll leave approval for someone from colab who has more knowledge of the API code than I do, but it is hard to see how this could hurt anything.
(Assuming "now" is in fact the correct time for "updated at", that is.)

denised (approved):

Issue #242, opened 2021-06-14
Corrupted model state (gremlin)


colleenskemp:
Time Estimate: 30 hours

Problem: Solutions are not calculating reliably and some solution calculations are not working. 

Steps to replicate: 
0) Clear redis in case the data is already cached
1) Create a new model
2) Add Solar Farms to the solution set
3) Edit Solar Farms
4) Change the adoption data source to conservative. 
5) Change growth to high
6) Calculate
7) Change adoption data source to high. 

Outcome: Nothing will change. The model will lock. 

Acceptance criteria: As a user, when I update the variables in any electricity generation solution, the model is able to calculate my solution output. 

Additional Notes: 
Changing the redis cache doesn't work. When the API code calls with solution code, something is failing.

colleenskemp:
Re-added some code that was removed. 
Added logging framework and made adjustments to make it not lock-up

PR #248, opened 2021-06-17
Corrupted model state (gremlin) #242

sutjin:
Found that in cases where data doesn't exist, exception object is set as a value that would be parsed as JSON, causing the code to lock up. This may be a cause to why result doesnt seem to change on the UI.

Now we will skip the missing data point and log what went wrong. Which will cause the UI to render below:
skipping_data_point

Also added a logger class that we can use for other purposes


brodavi (approved):
great thank you! :)

denised (approved):

Issue #257, opened 2021-06-24
Scenarios being overwritten when importing solutions scenario model


sutjin:
Problem: when re-running calculations, the `scenarios` key of scenario module is ovewritten to the base workbook instead if importing it from `solutions.<importname>/ac`

Step to replicate:
1. select a scenario and calculate
2. Change any of the fields and re-calculate

Outcome: Calculations are still being done accordingly, this is a backend issue

Acceptance Criteria: All scenarios under the  `solutions.<importname>/ac.` should be binded to the module instead of just 1

Additional Note:
* Affected Function: `factory_2.one_solution_scenarios` -> observe the `len(m.scenarios)`

denised:
When you say backend, do you mean code in /api or code in /solution?   I don't have that much knowledge of the code in /api, and a quick look tells me I need to know more about that do dig in here.

Happy to look at this together, though.

denised:
OK, actually have found the code in factory_2.py and I think I do understand the problem.  Let me make sure I do understand it:  the function `factory_2.one_solution_scenarios` is taking two arguments: a solution name and `j` which is presumably a json structure for override of the arguments for a single scenario.  But the code just overrides the entire `scenarios` data structure, instead of replacing just one.   I will fix this as a bug fix now, and replace the code with a more general override capability later.

PR #261, opened 2021-06-28
Fix for #257

denised:
Amend m.scenarios, not overwrite it.


sutjin (approved):
Great catch!

forgot that `m` points to the object and its being replaced

PR #269, opened 2021-07-02
making this into a library

sutjin:
Convert the project into a library.

able to be accessed via

`import solution`
`import model`
`import tools`

API related modules are not exported as libray as this project will only provide solutions/models in the future


brodavi (approved):
awesome!

denised (approved):

denised:
Approved and merged!

Issue #295, opened 2021-07-06
Solutions repo codebase cleanup


sutjin:
Need: Solutions repo is not used mainly for the models. As a Developer, I would like to remove any codebase that is not related to the solutions model

Acceptance criteria:
- remove codebase related to API
- remove dependencies not used by the model
- update README documentation

Note:
The API depends on 3 folders to be there:
- solution
- model
- tools

if any changes to the structure inside the folder are going to be changed/moved (unless adding new things), please communicate with the team as it may be breaking changes.

DentonGentry:
If I may note: https://github.com/ProjectDrawdown/solutions/tree/develop/limbo/dashboard is the code which generates https://projectdrawdown.github.io/solutions/ which I think could still be useful.

denised:
Yes, and the reason it is in limbo is because I would like to bring that back :-)
FWIW, however, the existing https://projectdrawdown.github.io/solutions/ site had not been updated in a long time, because of deployment errors.

-- 
  Denise Draper
  ***@***.***

PR #301, opened 2021-07-07
Cleanup step 1: delete migrated stuff

denised:


denised:
I'm going to divide #295 into pieces for easier review.  This first one just deletes things that are no longer necessary

sutjin (approved):

PR #303, opened 2021-07-09
Convert to openpyxl for most of the code

denised:
Converting to openpyxl for all of the code except the extraction tools.  Getting this working requires pinning the pandas version to 1.2.4.  Along the way got tests starting to run and fixed a bunch of minor issues (most of the __init__.py issues).


requirements.txt
```diff
@@ -1,4 +1,4 @@
 fair
 numpy-financial
-pandas==1.2
-xlrd==1.2.0
```
denised, line 4: @sutjin Switching from pandas 1.2 to pandas 1.2.4 --- I hope it doesn't cause a problem for you?

solution/afforestation/__init__.py
```diff
@@ -149,8 +149,7 @@ def __init__(self, scenario=None):
              'NOTE', 'NOTE', 'NOTE'],
             ['low_sd_mult', 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
             ['high_sd_mult', 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0]]
-        adconfig = pd.DataFrame(adconfig_list[1:], columns=adconfig_list[0],
```
denised, line 152: "dtype=np.object" is deprecated in pandas.  I could have replaced with dtype=object, but since dtype=object is the default for mixed-type dataframes, it didn't seem that this argument was necessary.
This same change occurs approximately 80 times below.

solution/test_factory.py
```diff
@@ -2,7 +2,7 @@
 
 import os
 import pandas as pd
-import factory
+from . import factory
```
denised, line 6: This change enables the tests to be run from the main directory.

glass-wing (approved):
Great to see the simplifications going in. The requirements.txt file was overly complex due to the api, alembic and postgres components.
I've been using pandas 1.2.4 for months now. It hasn't caused any problems for me.

Issue #267, opened 2021-07-01
Bug: change of size of operating cost breakout tables changed results


denised:
In  05988fd6, some tables in the operating cost calcs were truncated to a shorter length (comments indicate this was to improve performance).  This change directly broke the test suite, but more importantly changed some of the aggregated results.

denised:
d'oh, this was fixed already.

denised:
d'oh how confused can I get...

Issue #266, opened 2021-07-01
Hackathon prep: get tests working


denised:
Once the repo is split up, make sure we can run the test suite.

Issue #298, opened 2021-07-07
Jupyter notebook guide how to run the process during the hackathon


colleenskemp:
This will be instructions. 
This is relating to all the solutions in the notebook and is needed for hackathon prep.

Issue #297, opened 2021-07-07
Create Documentation supporting testing suite 


colleenskemp:
There are tests in a testing suite. Don't want to add more tests to the system. 

Documentation needs to be added that tells you how to run it.

colleenskemp:
Could making these tests be good hackathon work?

DentonGentry:
Other than, or instead of, https://github.com/ProjectDrawdown/solutions/blob/develop/tests/test_excel_integration.py ?

PR #305, opened 2021-07-12
Omnibus testing work and bug fixes

denised:
This is a catchall PR with a variety of bug fixes and improvements, focused on getting the tests clean.
Because it is so large, I'm going to list out the changes somewhat logically, together with which files they touched:

**Conversion of excel extraction code to openpyxl**  
* `tools/solution_xls_extract.py`:  lots of line-by-line changes; no functional changes.
* `tools/vma_xls_extract.py`: ditto.
* `tools/custom_adoption_xls_extract.py`:  deleted, since it was used nowhere, and the same function exists in solution_xls_extract.py  
* `tools/integration_test_xls_extract.py`:  this is the code that creates the expected.zip files.  It did not work on my Windows system since it relied on making a copy of the Excel file, which resets the "macro-enabled" bit, which defeats the whole purpose of using xlwings to operate it.  After trying unsuccessfully to find a workaround, I deleted that file and switched to a completely different implementation that uses VB macros to do the Excel work under the user's control and a python files to compose the results into a zip file:
    * `tools/CREATING_EXPECTED_ZIP.md`  explains how the whole process works
    * `tools/create_expected_zip.py` creates a zip in the right format file from csv files
    * `tools/export_csv.vb` VB code that runs an Excel and exports CSV files for all tabs, for all scenarios
* `tools/util.py`  Added convenience subroutines to make working with excel/openpyxl more consistent.
* `tools/tests`  updated to work with openpyxl; suppressed some tests that can't work yet.

**Add new extract_ghost tool**   (hackathon prep)
* `tools/extract_ghost.py`:  reverse the expected.zip to create a "ghost" of a spreadsheet for a single scenario.  I call it a ghost spreadsheet since has no macros, formulas, charts, graphs or even formatting, and only a subset of the pages.  But it is very helpful during debugging, especially since it is _exactly_ the data the test code is comparing against.

**Add Extraction Guide**  (hackathon prep)
The top-level files `Extraction_Guide.md` and `Extraction_Guide.ipynb` are added to help people go through the solution extraction process.  Should satisfy bug #298 

**Get (Most of the) Tests Running**  This is sufficient, I think, to close bug #266
* `tests/test_excel_integration.py`:  Fixed a bug that was comparing a dataframe and a series with different columns.
* `tools/health` and others.  Code that we are not currently using and would break tests has been moved to limbo/ directory.
* `solution/*/__init__.py`:  I missed a bunch of the type=np.object changes from the last checkin; this gets the remaining ones.
* `model/advanced_controls.py`: Code was dying on divide-by-zero.  I added some checks and returned zero in that case (which is consistent with the treatment of nan in other locations.  We should go back and verify the logic, however.  (I left a note in the code for that).
* `model/tests/test_vma.py`:  Suppressed a few tests (a) test functionality that we aren't using and (b) I think will be rewritten soon.
* `tox.ini`  Added suppression of openpyxl warnings that are spurious.

As of now, there remain about one dozen test failures in solutions (almost all in the same place), and less than half a dozen elsewhere.

**Added Testing Guide**   May satisfy #297 (Colleen will need to check)
* `TESTING.md`:  New file explaining the test system and instructions for how to run different kinds of tests.
* `CONTRIBUTING.md`:  Removed the now-redundant material from this file, and pointed to the new TESTING.md

**Bug Fixes**
* `model/operatingcost.py`: reverse part of change 05988fd that introduced a bug by truncating table sizes.  This fixes bug #267.
* `model/data_handler.py`: commit da187df098 added a regions parameter without a default, which was a backwards-incompatible change and broke the test.  I added a default of None and the logic to go with it.


CONTRIBUTING.md
```diff
@@ -30,30 +30,19 @@ similar to the following:
 
     Now you can work on the changes you have planned.  We encourage you to reach out for assistance as you work.  One way to do this easily is to create a Work-In-Progress (WIP) pull request (see the next section).
 
-    While you are working on your changes, test frequently to ensure you are not breaking the existing code. The project uses pytest, and invoking pytest from the root directory will run the entire test suite. There is a [YouTube video describing the different levels of testing](https://www.youtube.com/watch?v=K6P56qUkCrw).  
+    While you are working on your changes, test frequently to ensure you are not breaking the existing code. See [TESTING.md](TESTING.md) for guidance.   New code should include tests to cover the new functionality (and _must_ include test
```
denised, line 33: Moved this content to a new TESTING.md file

model/operatingcost.py
```diff
@@ -42,8 +42,8 @@ def annual_breakout(
     first_year = dd.CORE_START_YEAR
     last_year = report_end_year
     last_column = dd.CORE_END_YEAR
-    # last_row = 2139
-    breakout = pd.DataFrame(0, index=np.arange(first_year, last_year + 1),
+    last_row = 2139
```
denised, line 45: This is (and other changes in this file) is the fix to bug #267

tests/test_excel_integration.py
```diff
@@ -475,7 +475,7 @@ def verify_unit_adoption_calculations(obj, verify, include_regional_data=True, s
     # Carpool-RRSv1.1b-Jan2020.xlsm also catastrophic cancellation in multiple scenarios.
     baseline = obj.ua.soln_net_annual_funits_adopted().mean() * 1e-6
     s = obj.ua.soln_net_annual_funits_adopted().reset_index()
-    m = s.mask(s < baseline, other=True).where(s < baseline, other=False)
```
denised, line 478: This code simplification (repeated throughout this file) reduces the number of warnings appearing in the test run by about 800.   (There is still a warning for this line, however.)

tools/vma_xls_extract.py
```diff
@@ -282,8 +282,8 @@ def read_single_table(self, source_id_cell, sheetname, fixed_summary):
         col_names = list(df.columns)
         idx = 0
         skipcols = []
-        for c in range(16):
-            val = sheet.cell(row1, col1 + c).value
+        for c in range(col1, col1 + 16):
```
denised, line 285: Openpyxl translation for vma_xls_extract.py.

glass-wing (approved):
Good work Denise.

PR #320, opened 2021-07-19
polish multi_excel_sample and solution_xls_extract

denised:
Robustification of argument checking (accept str or Path args)
Multi_excel_sample does permissions checking before proceeding.


denised:
This is low risk and I'm doing testing for how merges etc. will work for hackathon prep, so I'm just going to merge it.

PR #353, opened 2021-07-24
WIP: Implement Household and Commercial Recycling in Python

alienau8:
Work in Progress


alienau8:
Found 'zh' value in './ca_pds_data/custom_pds_ad_PDS3_JUNE_2020_Postintegration.csv' and replaced by 0 as placeholder

alienau8:
The 'zh' value in 'Recycling-HH_C-v1.1b-MRG2020_PostIntegration' which was in the sheet 'Custom PDS Adoption' and cell B490 has now been replaced by a value calculated with the formula for the surrounding cells.

alienau8:
While the performing 'test_household_commercial_recycling' test we found an error with 'PDS2-68p2050-High Prognostication' scenario in the 'TAM Data' sheet in the third polynomial for the USA which can be found in the cells 'CI677:CM723'.
The constant is the expected date is off by 4.3043672314 compared to the actual.
For x^3: 0.7173945336 (weirdly 6 times less than the constant)
For x^2: 4.3043672044 (weirdly the exact same value)

Carson-Poe:
I've traced the issue to the poly_degree3_trend() function in Interpolation.py. I believe our np.polyfit() in python is calculating our coefficients differently then our excel function LINEST().

alexlod:
Here are some notes I made related to changes/observations that aren't captured in this PR:

---

CHANGE original excel sheet
    Sheet: Variable Meta-analysis
    Cell: K779
    Old value: kg CO2/t
    New value: Original Units

CHANGE original excel sheet
    Sheet: Custom PDS Adoption
    Cells: B56, B490,
    Old value: zh
    New value: copy-paste the formula from the adjacent cell

REMOVE the following scenarios. They're missing from the 'Custom PDS Adoption' sheet.
    - PDS1-65p2050-Medium Prognostication (Book Ed.1)
    - PDS1-65p2050-Medium Prognostication (Updated)

    Running them throws the following exception:

    ```
    TypeError                                 Traceback (most recent call last)
    /var/folders/c7/4m092rwx3xg7chj37j9xrjjr0000gn/T/ipykernel_1489/3586282394.py in <module>
        1 # Let's build the 2nd one
        2 
    ----> 3 myscenario = constructor(scenarios[7])

    ~/src/drawdown/solutions/solution/householdcommercialrecycling/__init__.py in __init__(self, scenario)
        310         ]
        311         for (i,rs) in enumerate(ca_pds_data_sources):
    --> 312             rs['include'] = (i in self.ac.soln_pds_adoption_scenarios_included)
        313         self.pds_ca = customadoption.CustomAdoption(data_sources=ca_pds_data_sources,
        314             soln_adoption_custom_name=self.ac.soln_pds_adoption_custom_name,

    TypeError: argument of type 'NoneType' is not iterable
```

FYI: https://github.com/alexlod/solutions/blob/develop/tools/CREATING_EXPECTED_ZIP.md -- the macros don't seem to work for mac. For example:
    `excellangID = Application.LocalizedLanguage` -- that doesn't work on Mac even though the comments say that line exists just for Mac.
    Instead, `excellangID = Application.LanguageSettings.LanguageID(msoLanguageIDUI)` works on Mac. But once I fixed that, I hit another
    issue likely caused by using a Mac.

**Comment by DentonGentry** [comment]
That usually means that the set of input points passed to Python interpolate.py doesn't actually match what is being used in Excel, even if it appears to. For example a common issue is when the Excel Advanced Controls lists 2018 as the base year, but is actually using 2014.

**Comment by Carson-Poe** [comment]
@DentonGentry I thought the same thing with regards to the 2014 date, but did some testing, and I'm really leaning into the functions computing the coefficients differently. I know screenshots are kind of taboo, but I think these highlight the problem-

Excel's LINEST():
image

And using the same input, Pythons np.polyfit(): 
image

The coefficients line up with the Assertion error.  

I think having 3 data points for this order 3 polynomial is asking a lot of precision from not a lot of data. 

We're open to any suggestions for the next steps.

**Comment by DentonGentry** [comment]
I have notes from one other time this happened, while importing AlternativeCements_v1.1c_IntegrationJune2020.xlsm. I think I ended up pulling in an Excel file of data to sidestep the problem altogether: https://github.com/ProjectDrawdown/solutions/blob/develop/solution/altcement/data.xlsx

The notes from the time are that sklearn's linear regression least squares implementation produced closer results to Excel, but still not within a floating point margin of error sufficient to produce the same result. My belief was that Excel's LINEST was also operating in an unstable region but just didn't say so.

The notes contain the hack to evaluate sklearn:
```
    def _linear_forecast(self, first_year, last_year, datapoints):
        """Interpolates a line between datapoints, and fills in adoption.
           first_year: an integer, the first year to interpolate data for.
           last_year: an integer, the last year to interpolate data for.
           datapoints: a Pandas DataFrame with two rows of adoption data, indexed by year.
             The columns are expected to be regions like 'World', 'EU', 'India', etc.
             There can be as many columns as desired, but the columns in datapoints
             must match the columns in adoption.
             The year+adoption data provide the X,Y coordinates for a line to interpolate.
        """
        adoption = pd.DataFrame(columns=datapoints.columns, dtype='float')
        year1 = list(datapoints.index)[0]
        for col in datapoints.columns:
            coeff = poly.polyfit(x=datapoints.index, y=datapoints[col], deg=1)
            for year in range(first_year, year1):
                adoption.loc[year, col] = poly.polyval(c=coeff, x=year)
            for year in range(year1, last_year + 1):
                adoption.loc[year, col] = poly.polyval(c=coeff, x=year)
        coeff = np.polyfit(x=datapoints.index, y=datapoints['World'], deg=1)
        print(f"np.polyfit: {coeff}")
        print(f"np.polyfit 2018: {np.polyval(p=coeff, x=2018):.15f}")
        coeff = poly.polyfit(x=datapoints.index, y=datapoints['World'], deg=1, rcond=1.e-25)
        print(f"poly.polyfit: {coeff}")
        print(f"poly.polyfit 2018: {poly.polyval(c=coeff, x=2018):.15f}")
        slope, intercept, _, _, _ = stats.linregress(x=datapoints.index, y=datapoints['World'])
        print(f"linregress: {slope} {intercept}")
        print(f"linregress 2018: {poly.polyval(c=[intercept, slope], x=2018):.15f}")
        reg = LinearRegression().fit(datapoints.index.to_numpy().reshape(-1, 1), datapoints['World'].to_numpy().reshape(-1, 1))
        coeff = reg.coef_[0][0]
        intercept = reg.intercept_[0]
        print(f"LinearRegression: {coeff:.15f} {intercept:.15f}")
        predict = reg.predict([[2018]])[0][0]
        print(f"LinearRegression 2018: {predict:.15f}")
        return adoption
```

Carson-Poe:
I mentioned this on the Slack Channel but I wanted to close the loop here as well:

It seems if we add one more row of data, our LINEST() and np.polyfit() functions compute the same coefficients. 

Excel:
image

Then the same with np.polyfit():
image

They match! 

So a possible fix is to add one more row of data to our excel spreadsheet.

Issue #347, opened 2021-07-20
Add utility conversions package.


denised:
This is a simple one :-)
On the Variable Meta-Analysis tab of any model, to the right starting at column P, are a series of tables expressing useful conversions.  In python we would like to have a utility module that would calculate these conversions, which could then be used in other calculations.  There are a couple of factors (`rrs.TERRAWATT_TO_KILOWATT` and `land.MHA_TO_HA`) that are expressed as constants, but we'd like to migrate these to functions (`def terrawatt_to_kilowatt(tw)`), put them in a single shared conversion module, and add the others found on the VMA tab:
* energy conversions
* volume conversions
* mass conversions
* distance conversions
* land area conversions
* inflation-adjusted-dollars

Note: CO2 and CO2-eq conversions are already implemented elsewhere.

ssmssam:
Hey, 
Is something like this what you are looking for https://github.com/ssmssam/solutions/blob/adding_utilitiies_module/model/conversions.py   

There are two functions replacing the rss and land constants, and have refactored the code to call these functions now instead. 

I've also created an energyConversion class, which has a lookup dictionary of the values from the first row and column in that table on the excel file. I used both the row and column to keep decimal rounding errors to a minimum so it could more closely replicate the values found in that excel sheet. 

The conversions are not always identical , e.g. in the code 1 twh = 859,845.2292000001 gcal
                                                                                in excel 1 twh = 859,845.228000 gcal 

If this is useful, I can generalise the class and make a metaclass which all the conversion classes will inherit from. Would just need to change; self.conversion_rates, self._base_name, self.accepted_names for each class

denised:
This looks really good @ssmssam --- I'll look into it in more detail as soon as I get a chance!

denised:
Sam I really like this approach --- much nicer than making tons of functions for every possible energy conversion, and leaves open the possibility of extensions later.
The metaclass approach should make the code very clean too.

Please continue!

ssmssam:
Great thank you :). I thought it might end up being a bit easier to use rather than having to look through all the functions to see which one is there. 
I will add unit tests then refactor to get the metaclass. Hope to be done with it all by end of tomorrow if that's ok.

ssmssam:
All done :)

denised:
Excellent work --- thank you!

PR #358, opened 2021-07-25
Adding utilitiies module

ssmssam:
#347  #WIP 
draft PR 
Need to generalise the class to a meta class and then create the other conversion tables using this.
Just want feedback on if this EnergyConversion class would be useful as it stands first

### DIFF
--- a/solution/afforestation/__init__.py
+++ b/solution/afforestation/__init__.py
@@ -383,7 +383,7 @@ class Scenario:
             soln_ref_new_iunits_reqd=self.ua.soln_ref_new_iunits_reqd(),
             conv_ref_new_iunits=self.ua.conv_ref_new_iunits(),
             conv_ref_first_cost_uses_tot_units=True,
-            fc_convert_iunit_factor=conversions.mha_to_ha())
+            fc_convert_iunit_factor=conversions.mha_to_ha)
 
         self.oc = operatingcost.OperatingCost(ac=self.ac,
             soln_net_annual_funits_adopted=soln_net_annual_funits_adopted,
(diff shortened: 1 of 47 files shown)


ssmssam:
metaclass added along with unit tests for all the conversion classes and functions

solution/afforestation/__init__.py
```diff
@@ -383,7 +383,7 @@ def __init__(self, scenario=None):
             soln_ref_new_iunits_reqd=self.ua.soln_ref_new_iunits_reqd(),
             conv_ref_new_iunits=self.ua.conv_ref_new_iunits(),
             conv_ref_first_cost_uses_tot_units=True,
-            fc_convert_iunit_factor=land.MHA_TO_HA)
+            fc_convert_iunit_factor=conversions.mha_to_ha())
```
denised, line 386: @ssmssam I was confused for a bit until I figured out how this is working.  I think it would be clearer if you explicitly put the default argument in.
````
fc_convert_iunit_factor=conversions.mha_to_ha(1)
````

It would be even more clear if the argument were changed like so:
````
fc_iunit_conversion=conversions.mha_to_ha
````
And then the First Cost code was changed to call the conversion function provided.   I'll leave it to you if you'd like to make this second improvement.

denised (changes requested):

denised (approved):
Very nice.  Very clean work, thank you!

solution/afforestation/__init__.py
ssmssam: I went down the second route, so FirstCost and OperatingCost will call the conversion factor if it is a function otherwise they will take its existing value

PR #359, opened 2021-07-25
Recycled Plastics Extraction

### LINKED ISSUE #287: NEW MODEL - Implement Recycled Plastic solution in python

cathbar:
Unit tests passing.

There were many issues with numbers being truncated and losing precision within the expected csv files. To overcome this, we switched the CSVs to be .xlsx files, which maintain their precision. These are then zipped and read through a `pd.read_excel()` in the unit tests. Because there are no file extensions in the expected files, it isn't easy to tell whether a user is using the legacy CSV format or the new xlsx format... So for now, we'll just try to read excel and then fall back to CSV if it fails.

Additionally, there are some flags that needed to be set in the `__init__.py` that we'd like some validation on from someone who understands exactly what those flags do.

resolves #287


cathbar:
Links to: Issue #287

denised:
Awesome work folks!    I'd like to separate out the test-result-generation into a separate PR, but take a few days off first :-)

cathbar:
We were curious if you'd want the XLSX stuff in a separate PR so that the recycled plastic solution is holding up other people who might need the XLSX stuff.

As much as I might like to hold the XLSX stuff hostage behind the solution extraction work...

denised:
Yes, that would be good.

-- 
  Denise Draper
  ***@***.***

cathbar:
I don't know how to refresh a github PR diff, but that `util.py` change is not an actual diff. It's already in the `hackathon` branch.

cathbar:
Tagging @everAspiring

cathbar:
Closing this because github is being a pain. Will reopen a new one.

PR #363, opened 2021-07-25
Csv to xlsx

### LINKED ISSUE #360: Change expected test files to use XLSX instead of CSV
There were many problems with numbers being truncated and losing precision in the csv files within `expected.zip`.  This loss in precision causes a whole lot of mismatched numbers in the unit tests. Using an excel workbook format instead of CSVs should preserve the precision. The unit test executor should be updated to handle both XLSX while still maintain backwards compatibility for CSVs.

cathbar:
resolves #360 

This PR will set the VBA to write an XLSX file and allow the test_excel_integration file to read it. It will fall back to reading CSV in the event of any exception. I'd normally try to put a more specific exception here, but I'm pretty sure we'd want to give CSV a try regardless of the exception.

### DIFF
--- a/tests/test_excel_integration.py
+++ b/tests/test_excel_integration.py
@@ -835,7 +835,6 @@ def excel_read_cell_any_scenario(zip_f, sheetname, cell):
         if sheetname in name:
             zip_csv_f = zip_f.open(name=name)
             (col, row) = cell_to_offsets(cell)
-
             try:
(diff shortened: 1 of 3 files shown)


cathbar:
Sorry about the whitespace problems in the VB script, but I can't get my IDE to play nice and honestly... that whitespace _is_ unnecessary.

denised:
No worries --- pushing now.

PR #369, opened 2021-07-28
hackathon july 2021 vma extraction bug fixes

anshulgoyal10:
In the process of extracting the Improved Cattle Feed model for July 2021 Hackathon, a few bugs/limitations in the extraction logic were found. This PR contains proposed fixes for them.

#### Change 1
In file `tools/vma_xls_extract.py`, modified line 419 to increased range of values looped over by variable `rows_to_next_table` from 200 to 220 to allow looping over tables that are slightly larger than 200 rows in the "Variable Meta-analysis" tab (as is the case for table 17 in ImprovedCattleFeed_Apr2021.xlsm which has 205 rows).

Without this change, extraction fails because function `_find_tables()` in tools/vma_xls_extract.py cannot find the start of VMA table 18 in ImprovedCattleFeed_Apr2021.xlsm because it is more than 200 rows away from the start of table 17.

#### Change 2
In file `tools/vma_xls_extract.py`, modified line 294 to increase range of values looped over by variable `r` from 140 to 210 to allow for tables that are larger than 140 rows in the "Variable Meta-analysis" tab (as is the case for table 17 in ImprovedCattleFeed_Apr2021.xlsm which has 205 rows).

Without this change, extraction fails because function `read_single_table()` in tools/vma_xls_extract.py stops reading table 17 in ImprovedCattleFeed_Apr2021.xlsm after the first 140 rows and throws the exception at line 316.

#### Change 3
In file `model/vma.py`, modified line 140 to have `pandas.read_csv()` automatically convert `'#DIV/0!'` and `'#REF!'` values to pandas NA so that pandas functions in the rest of the class don't throw exceptions when running math formulas on columns in `self.df` that are expected to only contain float values (no strings).

Without this change, extraction fails because in function `_discard_outliers()` in `model/vma.py`:
```python
def _discard_outliers(self):
    ...
    mean = df['Value'].astype('float64').mean(skipna=True)
    ...
```
the conversion `df['Value'].astype('float64')` throws an exception because some entries in `df['Value']` are `'#DIV/0!'` strings copied from the source Excel file.


denised:
@anshulgoyal10 I cherry-picked your bug-fix commits out of this PR into PR #372, so we don't need this PR any longer.  Thank you for those fixes!

PR #377, opened 2021-08-02
revamp CONTRIBUTING.md

denised:
In thinking about how to follow up with hackathon participants, I realized I need some different collaboration tools.  Something in between Slack (which is great for conversations with existing collaborators, but doesn't address the public) and a doc in the Repo itself (which is awkward for maintaining something that is supposed to be updated frequently).

This version of CONTRIBUTING.md proposes using the github wiki for public-facing project progress and planning --- I'm imagining something that is oriented towards outside participants who want to keep up to date with what is going on, or new people trying to get some context.  Something that will be updated on the order of monthly, not with every change.   (I'd take point on setting this up, but of course others are welcome to help ;-)

Meanwhile I also wanted to revamp the content away from a general "how to" (which is available elsewhere) and focus more on aspects that are unique to this repo.  One of which is incorporating the need for an NDA for some tasks, and providing a way to point people towards that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,63 +1,62 @@
 # Contributing to Project Drawdown Solutions
 
 Welcome to the Project Drawdown Solutions repository! We're excited that you're here and want to help
-with solutions to climate change.   These guidelines are designed to make it as easy as possible 
+with solutions to climate change.   These guidelines are intended to make it as easy as possible 
 to get involved. If you have any questions that aren't discussed below, please let us know by 
 [opening an issue](https://github.com/ProjectDrawdown/solutions/issues)!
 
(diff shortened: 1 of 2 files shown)


denised:
@colleenskemp I especially want your thoughts on this (both the process and the doc)

denised:
And check out the [wiki](https://github.com/ProjectDrawdown/solutions/wiki)  This is what I have in mind for it to look like.

Sunishchal (approved):
The changes here and wiki you created are excellent. The past activity section is especially nice; something that's updated regularly will provide helpful context for developers choosing what to work on next. 

Do we need to request a general NDA (for non-hackathon open source contributors) to link in this page? We'll definitely want to mention that it only applies for work that needs access to Excel files.

PR #391, opened 2021-08-10
GHG Accounting

krcolbert:
Hello, 

This push request is for updating the Project Drawdown Solutions model for the purpose of developing the greenhouse gas accounting method.

(1) Updated the greenhouse gas accounting method for the CO2 and CH4 modules. This way we can show results in both individual gas units and CO2-equivalent units. 
(2) Introduced a module for N2O. 
(3) Updated the Global Warming Potential (GWP) conversions in the emission factors. 
(4) Included a tutorial for how to use the new features with example solutions Improved Rice and Landfill Methane.

Warm regards, 
Kristina Colbert

### DIFF
--- a/KC_ghg_tutorial.py
+++ b/KC_ghg_tutorial.py
@@ -1,0 +1,454 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+
+# This document is meant to help teach how to use the Drawdown Solutions Python Model
+# and to demonstrate the use of the FaIR model.
+
+# Creator: Kristina Colbert
+# Last Updated: May 26, 2021
+
+# Table of Contents:
+    # Set up your Python Environment
+    # Basic Usage of the Drawdown Model
+    # Accessing the Results of a Scenario Run
+    # Using the Methane Module
+    # Multigas Emissions Reductions
+    # FaIR Tutorial
+    # Using the FaIR Model within Drawdown
+
+
+
+
+###########----############----############----############----############
+# SET UP YOUR PYTHON ENVIRONMENT
+
+# 1. Download the Drawdown Python Code from Github:
+   # https://github.com/ProjectDrawdown/solutions
+    git clone https://github.com/ProjectDrawdown/solutions.git
+    
+# This only downloads it to your computer, you'll need to tell Python that it's a
+# module that will be called. There are two options:
+  # a.) Everytime you open python, you will need to input on the console:
+      import sys
+      #sys.path.append('/Users/kristinacolbert/Box/Kristina Colbert/Python/solutions')
+      sys.path.append('/Users/kristinacolbert/Box/Kristina Colbert/GHG_Accounting/Python/Models/PD_Model/edited_copy/solutions')
+  # b.) Alternatively, you can make it permanent within the .bash_profile by 
+  # finding your .bash_profile (for me it's located in /Users/kristinacolbert/). 
+  # Make sure bash is the prescribed shell. For a Mac you can change this by:
+      chsh -s /bin/bash
+  # Once you have located your .bash_profile, open it, and add to the end:
+      export PYTHONPATH="/Users/kristinacolbert/Box/Kristina Colbert/Python/solutions"
+  # c.) If you are using Spyder as an IDE, you can go up to the menu bar and select python. Scroll down to 
+        PYTHONPATH manager. From there you can let it know where the solutions module path is located.
+
+# 2. Install the FaIR Simple Climate Model within Python:
+   # https://github.com/OMS-NetZero/FAIR
+    pip install fair
+    
+# Now you have the two most important modules installed. There may be a few more you need
+# and they are listed in the requirements.txt file within the solutions directory.
+
+# 3. Install required modules:
+    cd solutions
+    pip install -r requirements.txt
+"""       
+
+
+
+
+##########----############----############----############----############
+# BASIC USAGE OF THE DRAWDOWN MODEL
+
+
+# First thing to do is import the solutions module, and more specifically 
+# the script called "factory" which will be the "workhorse" of the python code
+import solution.factory
+
+# Running the following code will save all the included solutions to a dictionary
+# for an easy lookup.
+solutions = solution.factory.all_solutions_scenarios()
+
+# If you look at the saved variable "solutions", you'll find a list of solution nicknames
+# and each of the possible Project Drawdown Scenarios (PDS) associated with the solution
+
+# A full list can be overwhelming, so to focus on just one let's save Improved Rice
+solution1 = solution.factory.one_solution_scenarios("improvedrice")
+
+# For demonstration purposes, we are interested in running the one PDS scenario called
+# "PDS-86p2050-Plausible-Custom PDS-Avg" under the Improved Rice solution.
+# This is how we will run the single scenario:
+rice = solutions["improvedrice"][0](scenario="PDS-86p2050-Plausible-Custom PDS-Avg")
+
+# It should be super quick. So what's happening behind the scenes? 
+# Well, the line of code is calling on a specific script. The script is hidden at:
+# /solutions/solution/improvedrice/__init__.py 
+# About halfway down the "__init__.py" is a line that says "class Scenario:"
+# Below that line is all of the subscripts that are being run to produce our results!
+# If you're interested in checking out each subscript you can go into the 
+# /solutions/model/ directory to check each one out.
+
+
+
+
+##########----############----############----############----############
+# ACCESSING THE RESULTS OF A SCENARIO RUN
+
+# How about we check out a few cool results after running our scenario?
+# If you are on Spyder, which I recommend you use for your Python IDE software,
+# You can double-click on the last saved variable "rice" in the Variable explorer tab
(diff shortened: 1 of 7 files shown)


solution/improvedrice/__init__.py
```diff
@@ -372,6 +373,8 @@ def __init__(self, scenario=None):
 
         self.c2 = co2calcs.CO2Calcs(ac=self.ac,
             ch4_ppb_calculator=self.c4.ch4_ppb_calculator(),
+            ch4_megatons_avoided_or_reduced=self.c4.ch4_megatons_avoided_or_reduced(),
+            n2o_megatons_avoided_or_reduced=self.n2o.n2o_megatons_avoided_or_reduced(),
             soln_pds_net_grid_electricity_units_saved=self.ua.soln_pds_net_grid_electricity_units_saved(),
```
denised, line 378: Here you are referencing the self.n2o object which you have not yet created.  You need to move the block below above this one.

solution/landfillmethane/__init__.py
```diff
@@ -359,6 +360,8 @@ def __init__(self, scenario=None):
 
         self.c2 = co2calcs.CO2Calcs(ac=self.ac,
             ch4_ppb_calculator=self.c4.ch4_ppb_calculator(),
+            ch4_megatons_avoided_or_reduced=self.c4.ch4_megatons_avoided_or_reduced(),
+            n2o_megatons_avoided_or_reduced=self.n2o.n2o_megatons_avoided_or_reduced(),
             soln_pds_net_grid_electricity_units_saved=self.ua.soln_pds_net_grid_electricity_units_saved(),
```
denised, line 365: Same comment as for the other `__init__.py` file.

KC_ghg_tutorial.py
```diff
@@ -0,0 +1,454 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+
+# This document is meant to help teach how to use the Drawdown Solutions Python Model
+# and to demonstrate the use of the FaIR model.
+
+# Creator: Kristina Colbert
```
denised, line 8: There is a ton of fabulous content in this document --- and we need more documentation!
What I'd like to do with it is split it between the "Start_Here" jupyter notebook and the Solution_Model_FAQ file in the Documentation folder.
If you would like to do that work yourself, I would appreciate it, but I don't want that to be a blocker for getting this work done and included.  If you would rather leave that part to me, then we'll just move this file to the Documentation folder for now.

KC_ghg_tutorial.py
```diff
@@ -0,0 +1,454 @@
...
+
+
+
+###########----############----############----############----############
+# SET UP YOUR PYTHON ENVIRONMENT
+
+# 1. Download the Drawdown Python Code from Github:
+   # https://github.com/ProjectDrawdown/solutions
+    git clone https://github.com/ProjectDrawdown/solutions.git
+
```
denised, line 29: I want to avoid duplicated documentation for environment setup compared to what is in the README.md file.  It might we worth adding some of this information to that file, where it will improve readability.

KC_ghg_tutorial.py
```diff
@@ -0,0 +1,454 @@
...
+      chsh -s /bin/bash
+  # Once you have located your .bash_profile, open it, and add to the end:
+      export PYTHONPATH="/Users/kristinacolbert/Box/Kristina Colbert/Python/solutions"
+  # c.) If you are using Spyder as an IDE, you can go up to the menu bar and select python. Scroll down to 
+        PYTHONPATH manager. From there you can let it know where the solutions module path is located.
+
+# 2. Install the FaIR Simple Climate Model within Python:
+   # https://github.com/OMS-NetZero/FAIR
+    pip install fair
+
```
denised, line 48: FAIR will have to become a standard dependency for the solutions project, which will affect both the requirements.txt and environment.yml.  If it is okay with you, I'll update those.

KC_ghg_tutorial.py
```diff
@@ -0,0 +1,454 @@
...
+rice = solutions["improvedrice"][0](scenario="PDS-86p2050-Plausible-Custom PDS-Avg")
+
+# It should be super quick. So what's happening behind the scenes? 
+# Well, the line of code is calling on a specific script. The script is hidden at:
+# /solutions/solution/improvedrice/__init__.py 
+# About halfway down the "__init__.py" is a line that says "class Scenario:"
+# Below that line is all of the subscripts that are being run to produce our results!
+# If you're interested in checking out each subscript you can go into the 
+# /solutions/model/ directory to check each one out.
+
```
denised, line 90: Merging this with and expanding the Start_Here jupyter notebook would be ideal.

KC_ghg_tutorial.py
```diff
@@ -0,0 +1,454 @@
...
+    # oc: Operating Cost
+    # pds_ca: Custom Adoption Data for the Project Drawdown Scenario
+    # ref_ca: Custom Adoption Data for the Reference Scenario
+    # scenario: Naming Convention of the Scenario
+    # solution_category: LAND
+    # tla_per_region: Total Land Area for each Drawdown Region
+    # ua: Unit Adoption (a mix of functions to calculate annual adoption)
+    # units: Units for first cost, functional unit, implementation unit, operating cost
+    # vmas: Variable Meta Analysis Datasets
+
```
denised, line 124: And this would be a great addition to the FAQ!

KC_ghg_tutorial.py
```diff
@@ -0,0 +1,454 @@
...
+
+##########----############----############----############----############
+# FaIR TUTORIAL
+
+# FaIR (Finite-amplitude Impulse Response) Model is a simple climate-carbon cycle
+# model written in the python coding language. FaIR takes emissions of greenhouse 
+# gases, aerosol and ozone precursors, and converts these into atmospheric
+# concentrations, radiative forcing and temperature change.
+
```
denised, line 252: This section would be great as a standalone additional document in the Documentation folder.
I would suggest it should be a jupyter notebook, but If you wanted to leave it as a python file, I think that would be okay as well.

denised:
@krcolbert 

I was unable to load the improvedrice scenario due to the undefined self.no2 reference I mentioned in the comments.   Did it work for you?

Otherwise, I am pretty happy with what I see.   The main thing that has not been addressed is testing.  There are actually two parts to this: adding some unit tests for the new code (to the models/tests directory) and removing some of the existing tests of the solutions that will now break because the numbers they return are different.  The second one I will take care of once we have the scenarios implemented and working.  But I would appreciate it if you could create a test case or two for the new functionality --- I hope the existing tests are a reasonable guide, let me know if they aren't.

Other than that, please fix the scenario initialization, and add the initialization for the other affected scenarios.

I put a lot of comments in your wonderful documentation file --- I would like to see this better integrated into the rest of the code, but it is a much lower priority for getting this project accepted.

Thank you!

denised (changes requested):

solution/improvedrice/__init__.py
krcolbert: Fixed the undefined self.n2o reference. The n2o call function was out of order and needed to be placed above the CO2 functionality for the improved rice solution.

solution/landfillmethane/__init__.py
krcolbert: N2O is not a gas used in the landfill solution, so I removed all references to N2O there.

KC_ghg_tutorial.py
krcolbert: If you don't mind, can you make the changes you see fit for the Start_Here notebook and FAQ documentation? You have a better vision for what should be placed in there. But, you are more than welcome to use any content that I've created! Thanks!!

krcolbert:
Hi Denise,

Thank you for taking a look over the code and trying to run it. I did also end up having the same error as you about the N2O in the __init__.py. That was a mistake on my part. It is now corrected and the tutorial script will run for me. I think you have access to the second commit. Not sure how that all works - but let me know if you can't see it and we'll try to get it on here.

In the next commit, I will add the remaining multi-gas solutions. The current commit, only has edits to the improved rice and landfill methane solutions as two test examples. I can also put together a Jupiter Notebook for easier tutorial and test capabilities. 

Many thanks for you comments!
Kristina

KC_ghg_tutorial.py
denised: That makes sense.  I'll mark these comments as resolved.

denised:
Kristina --- That sounds great --- I look forward to seeing it.

-- 
  Denise Draper
  ***@***.***

krcolbert:
Hi Denise,

I've got a lot here that's been updated. And I'm pretty sure it will be ready to be merged. Everything is working on my end.

To give you a brief overview, I've added three Jupyter notebooks:
1.) Tutorial_Climate_Modeling
2.) Tutorial_GHG_Accounting
3.) Test_Multigas_Solutions

I also added some functionality to the Start_Here like we talked about doing.

And, there are several init.py files that are changed for the multi-gas solutions. You can test that they are all working with the Jupyter notebook called Test_Multigas_Solutions.

Let me know if you see anything that's not working out.
Kristina

denised:
@krcolbert This looks great!  I'm happy you did the documentation as well.  I've asked for signoff from Project Drawdown to accept this change into code base.

Issue #273, opened 2021-07-02
Implement Refrigerant Management - HFC Replacement in python


colleenskemp:
Problem: This solution is currently missing from the python code and needs to be added.

Current excel: Refrigerant Management - HFC Replacement	

Sector: Industry

Category: RRS

danielmk:
I started to look through the HFC Replacement excel. The good news is that for example the TAM extraction seems to be working with the standard extraction approach. However, there are many challenges. One of the most prominent issues: the "Adoption Data" sheet is entirely empty (and hidden in excel). I will start work on a specialized version of the solution_xls_extract.py module and my first step will be to extract adoption data from another sheet.

denised:
That sounds like you are already making a good start!  Let me know how it goes.
And, we can get help from the Project Drawdown researchers if you need it to understand what the spreadsheet is doing.

-- 
  Denise Draper
  ***@***.***

denised:
Oh that is funny --- there were some clauses I had put in AdvancedControls.py to protect from divide-by-zero, but I did that early on when I really didn't understand what was going on.   I just took them out because the places where they had been needed had all been bad scenario data, which has been fixed in other ways.

But from what I can see from a quick look at the spreadsheet, it seems that this model simply doesn't do any financial modeling at all.  So that means that AdvancedControls.py should not assume that those values will be set.  I will put the divide-by-zero clauses back in, which should fix this problem.

-- 
  Denise Draper
  ***@***.***

denised:
I've re-instated the divide-by-zero check.

-- 
  Denise Draper
  ***@***.***

johnpaulalex:
There's a bit of a typo in the excel which carried over to the expected.zip -- the PDS2 scenario is called "PDS2-82p2050-Median " (with a space at the end).  For general sanity (and to match the PDS2 variable in the solution's __init__.py), I removed the trailing space from the PDS2-82p2050-Median/ subdir and from the two references in PDS2-82p2050-Median/ScenarioRecord (also in that zip file).  Here's the resulting
[expected.zip](https://github.com/ProjectDrawdown/solutions/files/7247818/expected.zip).

PR #397, opened 2021-08-14
[WIP] HFCReplacement Model as in #273

danielmk:
Port of the HFCReplacement model as discussed in #273. It is now possible to construct a scenario without error. This is work-in-progress because the testing fails. Working through the tests will take me some time so I decided to open this WIP PR. I suspect that the lack of financial analysis in this model is not handled as expected by the test suite and I am working to verify.


denised:
You are correct that the test suite does not recognize models that do not have a financial model.  You will need to add the corresponding test groups ('First Cost') etc. to TEST_SKIP.

danielmk:
Thanks for the `TEST_SKIP` hint. Skipping for example `'Operating Cost'` will be necessary but is not sufficient to make the test pass, because the error occurs before the `TEST_SKIP` mechanic works. Specifically, when `RRS_solution_verify_list` calls `verify_operating_cost`, it raises `AssertionError: Cannot have a lifetime replacement of 0 and non-zero operating costs` and `TEST_SKIP` does not avoid this.

My instinct would be to pass the TEST_SKIP flag to `RRS_solution_verify_list` and introduce conditionals to skip `verify_operating_cost`. Would that be a good strategy or should I try to find another workaround?

denised:
That would be wonderful, but it would also be a lot of work.
The way the testing is written, as you have noticed, all the test procedures are run before any of the test comparisons take place.  When I added the test_skip and test_only capabilities, I inserted them in the comparison part of the testing because it was much easier to do --- there's a convenient description to match against, and the code is only in one place.

Ideally we would use the flags to control the tests before they are executed --- it would make sense and make the tests run faster.  But it a pretty laborious task that would require a lot of refactoring.

So I have two -- or even three -- alternative suggestions.

The quickest is to temporarily override the test code directly (comment out the stuff that is causing you problems) in order to verify that the rest of the testing runs and gives the right results, but then check in the project with test failure.  (It would not be the only solution that currently fails tests, but appears to run just fine.)

The second would be to modify the test code to not throw an exception where it does --- perhaps to add specific flag to mark some solutions as having no financial modeling --- and then let the tests behave accordingly.  This is really just a specific version of the test_skip code for this particular case.  It would be a correct approach, but it feels like too much work for too little gain to me.

The third option is:  we really need a different kind of acceptance testing that is based on the final outputs (the so-called "key results").  Ideally every solution will pass the "deep" testing that we are currently doing, but they really must pass the "key result" testing --- but currently we don't even define those key results, let alone test them. 

My recommendation is that you take the first option, but if you are up for it, I would dearly love to see the third option happen, perhaps as a follow-on project.

-- 
  Denise Draper
  ***@***.***

danielmk:
Thank you for the detailed options. I agree that the first option is the best for this current PR and I'll move forward with it. I'd be happy to talk about the third option and implement key results next.

denised:
@danielmk Did you finish testing on this --- is it ready to go?

danielmk:
Besides the finance related errors that I described the test suite also turned up divergence between expected results and Python implementation for CO2 Calcs. I am tracking this down right now. Once I have correct CO2 calculations I'll let you know and the PR should be ready to go.

danielmk:
Tests pass now but I have to skip scenarios 1 and 2. Having the 0th scenario pass tests is a good starting point for this solution. The only change I made in the test suite was in 4de179f where I added a check for this particular solution
`if not obj.name == 'Refrigerant Management - HFC Replacement':`
From my side the solution is now ready for merge.

PR #398, opened 2021-08-15
Add Recycled Metals and new Adoption quirk

denised:


model/adoptiondata.py
```diff
@@ -33,11 +34,16 @@ def __init__(self, ac, data_sources, adconfig, main_includes_regional=None):
                'Latin America', 'China', 'India', 'EU', 'USA'
              main_includes_regional: boolean of whether the global min/max/sd should include
                data from the primary regions.
+
+            Quirks parameters:
+                groups_include_hundred_percent.  Some models included the 100% / maximum case as
```
DentonGentry, line 39: This sounds really familiar.
 https://docs.google.com/document/d/1OiKg3_OOGjYOUdnHTQuZggsko5n31qv_YV4h77E3LHk/edit (search for STDEV.P) shows two cases where one of the adoption columns was being included in the mean but not the stddev, which seemed nonsensical. I modified the spreadsheets to include the column in the STDEV before importing and generating the test_excel_integration ZIP files for those solutions.
 
I don't see any mention in that doc about the 100% case in STDEV, so I guess even though it sounds familiar that it has not been run into before.

The doc does mention a number of cases where the spreadsheet name was changed to "100% Case" but a formula was still looking for the former "100% RES Case"

denised:
The distinction is very clear: the Energy Models do not consider it, all other models do.  This is the Excel formulas in the two cases:

Energy 

=IF($B$38=$C$44,IFERROR(STDEV.P(C53:F53),""),
IF($B$38=$G$44,IFERROR(STDEV.P(G53:K53),""),
IF($B$38=$L$44,IFERROR(STDEV.P(L53:Q53),""),
IF(Y53="","",STDEV.P(C53:V53)))))

Other

=IF($B$37=$C$44,IFERROR(STDEV.P(C46:E46),""),
IF($B$37=$F$44,IFERROR(STDEV.P(F46:H46),""),
IF($B$37=$I$44,IFERROR(STDEV.P(I46:K46),""),
IF($B$37=$L$44,IFERROR(STDEV.P(L46:M46),""),
IF(T46="","",STDEV.P(C46:M46))))))

Until there was a 100% case with multiple columns, one of which was empty, the two different interpretations nicely dovetailed.

-- 
  Denise Draper
  ***@***.***

PR #405, opened 2021-08-22
Factory.py supports dynamic scenario without a hack

denised:
This is step one of supporting uploaded adoptions and tams.

One problem with the previous factory_2.py hack is that it overwrote a global data structure, which meant that you couldn't revert to the old definition of the hacked scenario.  I ran into that issue immediately when testing overloading tams.

I have also renamed the methods, since I found  `one_solution_scenarios` cumbersome to explain and use.
It is replaced by two methods:  `list_scenarios` lists all the scenarios, and `load_scenario` loads a single scenario, which may either be a string title (as before), or an AdvancedControl object, to create a completely custom scenario.

I haven't updated the extraction code or all the documentation yet; will do that the dynamic updates as a whole.

@sutjin @glass-wing I will want an actual review from both of you.


denised:
The vast majority of the changes in the commit are boilerplate.   What you want to review is factory.py and _one_ of the `__init__.py` files, which I will try to mark with comments.

solution/afforestation/__init__.py
```diff
@@ -129,10 +129,13 @@ class Scenario(scenario.Scenario):
...
-            scenario = list(scenarios.keys())[0]
-        self.scenario = scenario
-        self.ac = scenarios[scenario]
+       
+        if isinstance(scenario, ac.AdvancedControls):
+            self.scenario = scenario.name
+            self.ac = scenario
+        else:
+            self.scenario = scenario or PDS2
+            self.ac = scenarios[self.scenario]
```
denised, line 138: This identical change is done to all `__init__.py` files.

solution/factory.py
```diff
@@ -14,37 +14,32 @@ def all_solutions():
     # The above test is sufficient, but to be nice, allow for two more ways to make something not a solution
     return [ name for name in candidates if not name.startswith('_') and not name.startswith('test') ]
 
+def list_scenarios(solution):
+    """Return a list of scenarios for this solution"""
+    m = _load_module(solution)
+    return list(m.scenarios.keys())
```
denised, line 20: `one_solution_scenarios` returned a (constructor, scenario-list) tuple.  The advantage of doing things this way was that it cached them with `lru_cache` which meant that the dynamically loaded module could be efficiently re-used---as long as you remembered to use the same constructor.  In place of this, I now have `list_scenarios` and `load_scenario` both call `_load_module` under the hood, which directly caches the module itself, enabling the same and arguably better re-use.

solution/factory.py
```diff
@@ -14,37 +14,32 @@ def all_solutions():
...
+
+def load_scenario(solution, scenario=None):
+    """Load a scenario for the requested solution.  Scenario may be one of the following:
+     * None (the default): return the PDS2 scenario for this solution
+     * `PDS`, `PDS2` or `PDS3`:  get the most recent scenario of the requested type
+     * a scenario name:  load the scenario with that name
+     * an Advanced_Controls object:  load a completely custom scenario based on the data in the object"""
+    m = _load_module(solution)
+    if scenario in ['PDS1','PDS2','PDS3']:
```
denised, line 29: I have also inlined the functionality that I previously had put in a separate method to make it easy to access PDS1, PDS2 and PDS3 scenarios without ever even having to look at scenario names.

glass-wing (approved):
Great work. Looks good to me.

denised:
@sutjin I'm going to go ahead and pull this now, so I can get the tam upload one in next.

PR #409, opened 2021-08-24
Improve start here

fsboehme:
added a couple examples as I'm figuring out how to use this. perhaps someone who knows the project better can add a few more examples to help newbies such as myself to find their way around and become productive contributors?


denised:
@fsboehme I totally agree we need more here!  But I suggest holding off on this work, because I have another PR that also has changes to Start_Here, and dealing with conflict resolution in Jupyter Notebooks is a pain.

The other PR (#391) is expected to be going in later this week.

fsboehme:
Great! I'll be looking out for that! Thank you, @denised !

Issue #426, opened 2021-08-30
Solution models in python produce key results


colleenskemp:
Description: 

Key results are results that appear at the top of the advanced control tab and summarize the output of the models)

It is these key results that we can compare against the outputs from the excel spreadsheets and evaluate if we are meeting the key requirement of our python model producing consistent results. 

Acceptance criteria: 
All solution models produce key results.

denised:
Issue #304 is duplicate, but has some additional information that is still useful.

Issue #427, opened 2021-08-30
"Start here" notebook shows off the key results.


colleenskemp:
Description: The start here notebook should display the key results in a highly visible way. 

This is where a user starts when using the models.

This relates to #425 and #426

Issue #425, opened 2021-08-30
Create new tests for key results


colleenskemp:
Description: 

This ticket is focused: Need tests that test the key results. 

Key results are results that appear at the top of the advanced control tab and summarize the output of the models)  

Acceptance criteria need to be based on key results. 

Current problem: Right now there are a lot of test failures that do not relate to key results. It is the key results that help us evaluate if we are meeting the key requirements of the project. 

It relates to creating to the following (tracked in separate tickets):
* This ticket which has been closed as duplicate: test polishing #374
* The models have to produce key results
*  The start here notebook should show off the key results. This is where you start when you use the models.

colleenskemp:
@denised - This ticket is intended for Daniel, following the conversation we had last week. Could you add additional context and detail in this ticket for him? We can also discuss further detail during his onboarding later today.

denised:
Sorry Colleen --- the other issue that is duplicate is #378, not #374.

Issue #378, opened 2021-08-02
Tests to focus on key outputs


denised:
Currently our result testing tests all kinds of intermediate results against Excel, even when those are not used or do not have a significant impact on the actual results.  This is an important tool to validate deep consistency, but it is (a) very slow to run and (b) does not help us prioritize work that is having the most impact on the results.

We would like to have a separate test that would specifically test _outputs_ of the model for consistency with the Excel results.

Note issue this may be undertaken on its own, but it would make sense to combine it with #304

denised:
Closed as duplicate of #425

PR #428, opened 2021-09-01
[WIP] Add key results and test them

danielmk:
This is work in progress to port key results, as they are found in the Advanced Control sheet in the excel implementation, into Python. Issues related to this PR are #427 #426 #425 #304 and #378

All six key results are now implemented in model/scenario.py as part of the RRSScenario class. I am focusing on the RRS case for now and worry about generalizing to LandScenario later.

My next step is to write a test suite for the key results. The notebook Key_Results.ipynb shows current usage of key results, parts of it can be merge into Start_Here.ipynb later.

Suggestions regarding implementation and usage welcome.


denised:
Thanks Daniel!
I want to alert @glass-wing to this work --- Neil, please take a look at this too.

danielmk:
I confirmed that the current implementation works for bikeinfrastructure

However, it fails for solutions, for example altcement with this error message:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~\AppData\Local\Temp/ipykernel_16296/2792129031.py in <module>
      2 import pandas as pd
      3 pds1 = factory.load_scenario("altcement", "PDS1")
----> 4 pds1.key_results()

~\repos\solutions\model\scenario.py in key_results(self, year, region)
    113 
    114     def key_results(self, year=2050, region='World'):
--> 115         return {'implementation_unit_adoption_increase': self.implementation_unit_adoption_increase(year=year),
    116                 'functional_unit_adoption_increase': self.functional_unit_adoption_increase(year=year),
    117                 'marginal_first_cost': self.marginal_first_cost(year=year),

~\repos\solutions\model\scenario.py in implementation_unit_adoption_increase(self, year, region)
    121 
    122     def implementation_unit_adoption_increase(self, year=2050, region='World'):
--> 123         return (self.pds_ca.adoption_data_per_region().loc[year][region] / self.ac.soln_avg_annual_use - 
    124             self.ref_ca.adoption_data_per_region().loc[year][region] / self.ac.soln_avg_annual_use)
    125 

~\repos\solutions\model\customadoption.py in adoption_data_per_region(self)
    264     def adoption_data_per_region(self):
    265         """ Return a dataframe of adoption data, one column per region. """
--> 266         if self.soln_adoption_custom_name.startswith('Average of All Custom'):
    267             (result, _, _) = self._avg_high_low()
    268         elif self.soln_adoption_custom_name.startswith('High of All Custom'):

AttributeError: 'NoneType' object has no attribute 'startswith'
```

I had a hard time fixing this error. Setting `soln_adoption_custom_name = ''` instead of None produced another error. I also don't understand why it is None in the first place. For bikeinfrastructure it is `'Drawdown Book Edition 1 Scenario 1'` as I would expect from the excel. Any help appreciated.

denised:
@danielmk  The problem you are having is that not all models use custom adoption.  The class HelperTables (and the tab HelperTables in the Excel) is where the final selection of adoption is done.  So when you look for adoption, you want to look at `HelperTables.soln_pds_funits_adopted()`  or `HelperTables.soln_ref_funits_adopted()`.

Helper Tables is an *awful* name, and we will change it one day.

denised:
@danielmk I have just pulled in the first of the oceans models.  Neil (@glass-wing) is using entirely custom code for these models, not based on the existing code at all.  One of the things I want the key results to do is work over both types of models.   (We will also need to make the factory code work over both models, which I will take a look at soon.)   Please ask Neil for any help you need to get the results you need out of the oceans models.

danielmk:
Thank you Denise, your tip regarding Helper Table was very helpful. I now check for the existence of pds_ca and ref_ca and refer to Helper Table when necessary.

I now implemented testing for key results. I tried the tests on a variety of RRS solutions and many of them pass the test. However, the numbers diverge for some scenarios in some solutions. For example, bikeinfrastructure PDS2 and bioplastic PDS2 key_results are not as expected. Notable, PDS1 passes tests in both cases. I was so far not unable to trace this divergence.

I will next check if I can generalize the implementation to land solutions and afterwards move on to Neil's ocean model.

denised:
A number of solutions are failing the existing tests (bikeinfrastructure is one of them), so it is not surprising that you will find failures in the new tests.   It is also normal for a solution that one scenario may pass while another fails.

These are the solutions that currently fail the existing tests.  My hope is that some of them will actually *pass *key results testing, but I certainly don't expect all of them to.

FAILED solution/afforestation 
FAILED solution/bamboo
FAILED solution/bikeinfrastructure
FAILED solution/biogas
FAILED solution/bioplastic
FAILED solution/composting
FAILED solution/conservationagriculture
FAILED solution/electricbikes
FAILED solution/forestprotection
FAILED solution/greenroofs
FAILED solution/hcrecycling
FAILED solution/improvedcookstoves
FAILED solution/improvedrice
FAILED solution/landfillmethane
FAILED solution/masstransit
FAILED solution/methaneleak
FAILED solution/peatlands
FAILED solution/perennialbioenergy
FAILED solution/riceintensification
FAILED solution/ships
FAILED solution/solarhotwater
FAILED solution/trains
FAILED solution/walkablecities
FAILED solution/wastetoenergy

-- 
  Denise Draper
  ***@***.***

danielmk:
`key_results` now works for RRS and LAND scenarios and testing the key_results is now part of the standard test suite. It looks promising as some scenarios (for example bamboo) pass key_results testing but fail the more detailed tests as we hoped for. 
Key_Results.ipynb demonstrates the usage of `key_results`.

Regarding the ocean models I talked to @glass-wing today. His implementation of the ocean models are very different from the other model types. That also goes for the testing side (with some vast improvements of his from what I saw so far). I guess the goal would be to implement the .key_results interface for the ocean model that is more suited to the factoring of these models. I will be looking for a good way of implementing and possibly testing key results in the ocean models but I think that might be worth another PR because it will involve changes outside of scenario.py

@denised could you weight in on whether the ocean implementation warrants another PR or you prefer to have it here?

danielmk:
ocean_solution.py is well designed so I could implement key_results for ocean much faster than expected.

denised:
Excellent!  Great progress!

I would like the same key_results method to work with both Land/RRS models and the Ocean models.  It does not need to be integrated into oceans testing, however — I think we can leave Neil’s tests as they are.

Whether you do it in this PR or as a separate one is up to you.  Perhaps one reason to go with what you have now is that I have a rather large PR on the way in the next day or so, and it might be better for yours to go first.  Let me know.

Thanks!

-- 
  Denise Draper
  ***@***.***

danielmk:
@denised key_results() are now working for all model types and the PR is ready to merge from my side.

model/scenario.py
```diff
@@ -35,6 +35,115 @@ def set_ref_adoption(self):
     def set_pds_adoption(self):
         pass
 
+    def key_results(self, year=2050, region='World'):
+        if self.solution_category == self.solution_category.REDUCTION or self.solution_category == self.solution_category.REPLACEMENT:
```
denised, line 39: I think rather than use the solution_category field, it would be cleaner to use the inheritance:  put the REDUCTION and REPLACEMENT code in RRSScenario, and the LAND code LandScenario.  But this can be done at a later time.

model/scenario.py
```diff
@@ -90,8 +199,79 @@ def set_tam(self, config_values=None, **args):
             tam_ref_data_sources=self.tam_ref_data_sources,
             tam_pds_data_sources=self.tam_pds_data_sources,
             **args)
+    """
```
denised, line 202: I think this is leftover and can be removed, yes?

denised (approved):
A few small nits can be done in a subsequent checkin

denised:
Thank you @danielmk !

denised:
This closes #426 and #425

PR #430, opened 2021-09-04
tam and sma

denised:
introducing sma (Series Meta-Analysis), future ancestor of tam and adoption.
Moving tam initialization to tam.py


denised:
These changes were made to support integrations, but they will prove useful elsewhere too.

PR #431, opened 2021-09-04
Doc that Windows Excel is needed for creating expected.zip

johnpaulalex:
… do.

### DIFF
--- a/tools/Extraction_Guide.ipynb
+++ b/tools/Extraction_Guide.ipynb
@@ -110,7 +110,7 @@
     "* One set of PDS1, PDS2 and PDS3 scenarios that are recent.  These will usually have names with PDS1/2/3 in the title, or they may be have the labels \"Plausible\", \"Drawdown\" and \"Optimum\" instead (which mean the same thing).  We prefer the most recent set of three.\r\n",
     "* If there are a set of three scenarios labeled something like \"Book Version\", keep those as well.\r\n",
     "\r\n",
-    "Make a git commit that has all the generated scenarios in it, then delete scenarios that we don't need.\r\n",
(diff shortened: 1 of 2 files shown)


tools/Extraction_Guide.ipynb
```diff
@@ -183,7 +183,7 @@
    "source": [
     "## Create Test Results\n",
     "\n",
-    "**This step requires the Excel application, and thus only can be run on Windows or Mac.**\n",
+    "**If you don't have Windows Excel (Mac Excel has trouble with some of the macros), ask someone to do this process for you.**\n",
```
DentonGentry, line 186: Many of the existing expected.zip files were created using MacOS Excel 2016. 

Is the macro issue you've noted in LoadScenario_Click and sFoV? A number of the Excel files had language handling which broke with Mac Excel, the import steps include how to fix it to allow Mac Excel to work. https://docs.google.com/document/d/1OiKg3_OOGjYOUdnHTQuZggsko5n31qv_YV4h77E3LHk/edit

johnpaulalex: Thanks Denton, yeah I did those. The new issue was that the macro refused to write the csvs for the commercialglass solution, saying the output dir was read-only. But if I was able to save the CSVs by hand to that same dir, no problem. But I'm just the messenger here about policy - Denise said there were other problems with the Mac and asked me to modify this text accordingly.

denised: Its a different macro, Denton.  When I first tried to create expected.zip, the code at that time would not work on my windows machine, so I tried a different approach.  My hope was to get something that would work on both, but it seems I did not succeed.

denised (approved):

PR #435, opened 2021-09-10
key results refactoring

danielmk:
This PR includes several changes to key results that we discussed via email @denised 

Key results are now refactored into `RRSScenario` and `LandScenario`, `ht` is always used for adoption data and I changed from numpy `assert_almost_equal` to pytest.approx for testing.

Also `key_results()` is now called `get_key_results()`

### DIFF
--- a/model/scenario.py
+++ b/model/scenario.py
@@ -144,14 +144,17 @@ class Scenario:
     # Key Results
 
     def get_key_results(self, year=2050, region='World'):
-        if self.solution_category == self.solution_category.REDUCTION or self.solution_category == self.solution_category.REPLACEMENT:
+        if (self.solution_category == self.solution_category.REDUCTION or 
+            self.solution_category == self.solution_category.REPLACEMENT or
+            isinstance(self, RRSScenario)):
             return {'implementation_unit_adoption_increase': self.implementation_unit_adoption_increase(year=year),
                     'functional_unit_adoption_increase': self.functional_unit_adoption_increase(year=year),
                     'marginal_first_cost': self.marginal_first_cost(year=year),
                     'net_operating_savings': self.net_operating_savings(year=year),
                     'lifetime_operating_savings': self.lifetime_operating_savings(),
                     'cumulative_emissions_reduced': self.cumulative_emissions_reduced(year=year, region=region)}
-        elif self.solution_category == self.solution_category.LAND:
+        elif (self.solution_category == self.solution_category.LAND or
+              isinstance(self,LandScenario)):
             return {'adoption_unit_increase': self.adoption_unit_increase_LAND(year=year),
                     'marginal_first_cost': self.marginal_first_cost(year=year),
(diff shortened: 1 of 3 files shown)


Business-Wizard (approved):
LGTM - general review

model/scenario.py
```diff
@@ -311,6 +255,23 @@ class LandScenario(Scenario):
     def adoption_limit(self):
         return self.tla_per_region
 
+    def get_key_results(self, year=2050, region='World'):
+        return {'adoption_unit_increase': self.adoption_unit_increase(year=year),
```
denised, line 259: @danielmk: I suggest using the same key 'functional_unit_adoption_increase' for both RRS and Land, so you can write code that will work for both of them without adding a conditional.

tools/expected_result_tester.py
```diff
@@ -1067,21 +1067,19 @@ def key_results_tester(solution_name, expected_filename, scenario_skip=None):
...
-                aae(key_results['net_operating_savings'], float(df_expected.loc[3, 2]), decimal=desired_precision)
-                aae(key_results['lifetime_operating_savings'], float(df_expected.loc[3, 3]), decimal=desired_precision)
-                aae(key_results['cumulative_emissions_reduced'], float(df_expected.loc[3, 4]), decimal=desired_precision)
-                aae(key_results['total_additional_co2eq_sequestered'], float(df_expected.loc[3, 5]), decimal=desired_precision)
+                assert(key_results['adoption_unit_increase'] == pytest.approx(float(df_expected.loc[3, 0])))
+                assert(key_results['marginal_first_cost'] == pytest.approx(float(df_expected.loc[3, 1])))
+                assert(key_results['net_operating_savings'] == pytest.approx(float(df_expected.loc[3, 2])))
+                assert(key_results['lifetime_operating_savings'] == pytest.approx(float(df_expected.loc[3, 3])))
+                assert(key_results['cumulative_emissions_reduced'] == pytest.approx(float(df_expected.loc[3, 4])))
+                assert(key_results['total_additional_co2eq_sequestered'] == pytest.approx(float(df_expected.loc[3, 5])))
```
denised, line 1078: @danielmk It would be nice to see what the two numbers are in the test output when the test fails.

denised (approved):
@danielmk: Very nice; makes the code easier to read as well!.  I left comments for a couple of nits.
Do you want to continue to push changes in to this PR, or should I merge this and you will continue work on a new one?

danielmk:
I think merging this would be good. I will open a new PR soon that addresses some of the key_results testing failures and then I would make separate ones to address your comments.

model/scenario.py
denised: Or I like this idea even better: use the 'unit_adoption_increase' key in both RRS and Land, and let RRS have the "implementation adoption" as an extra key.

denised:
@danielmk Done.

PR #437, opened 2021-09-11
added comment with script for installing library

Business-Wizard:


denised:
@Business-Wizard which tests show an error, and what platform are you running them on?
And what is the purpose of the comment for fair?

Business-Wizard:
While setting up my env to get started adding value, I realized not all of the libraries installed properly.  It seems fair is not on the conda-forge repository, so I did the research and found the needed script to install it for new users. 😊

Business-Wizard:
Running tests on an Ubuntu 20.04. 
Haven't figured out how to get the yaml library working for the tests yet - that is the only remaining 8 tests throwing errors.

denised:
OK.  Requirements.txt is meant to be used with pip, which does know how to install fair.   (And we use pip in to install fair in the conda file environment.yml as well.)   So I think we're covered on that one.

Business-Wizard:
Ah, neat - I'll spend some time figuring how to properly use the environment file later and see if that doesn't just fix it. 
Hope it resolved the yaml errors too, and I can remove this PR. 😎

PR #438, opened 2021-09-11
General cleaning and refactorings

Business-Wizard:
Made easily possible by Sourcery extension.

### DIFF
--- a/solution/factory.py
+++ b/solution/factory.py
@@ -44,7 +44,4 @@ def _load_module(solution):
     return m
 
 def all_solutions_scenarios():
-    everything = {}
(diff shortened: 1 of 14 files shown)


denised:
Thank you for your efforts on this @Business-Wizard.  I like some of these changes, but not all of them.  Unfortunately github doesn't have a line item veto :-)   I'm going to reject the PR, but keep the list of changes and make some of them when I get around to modifying those files.

Business-Wizard:
Haha, well they do allow code block comments and suggestions. (So helpful!)

Can you point out what portions you didn't like so we can discuss and adapt for future PRs? @denised

Issue #317, opened 2021-07-18
Current Model Test Failures


denised:
The following models currently have test failures, all on the same unit adoption test:
bikeinfrastructure, buildingautomation, composting, concentratedsolar, electricbikes, greenroofs, masstransit, solarhotwater, walkablecities.  Since they previously worked, and are all failing at the same location, it is probably a recent bug (likely having to do with replacing xlrd with openpyxl).

denised:
The current goal is to look into test failures and see if some can be easily fixed.  This should be a limited time effort of a week to a week and a half.

danielmk:
In bikeinfrastructure the issue is in `pds.ua.soln_pds_cumulative_funits()`
The values for World are correct. In the other regions the expected values are all 0, because those are undefined in the excel. However, Python puts values for the other regions starting in the year 2018. Those values come from the `ref_base_adoption` variable in the ac .json files.
Not sure what's the best fix here.

denised:
Usually it turns out that different spreadsheets did things differently, and the extractor code tries to match that by adding in what I call quirks parameters (like bug_cfunits_double_count).  But the extractor may get it wrong and set the wrong value.  Or in some cases, this is a new quirk that neither the extractor nor the code know how to handle.

If there are existing relevant quirks parameters, looking at those and trying different values may help.
Or, to see if this is a new quirk, the multi-extract tool can be very useful.  It will show you the same bit of Excel from multiple spreadsheets so you can quickly compare the formulas and see if they differ between different spreadsheets.  If they do, then we may need to add a new quirks parameter to handle that.  (I think you have access to the entire set of spreadsheets -- if not, let me know.)

This particular error ('Q135:AA181') is actually the error we see in quite a few of the failures, so figuring this out for one solution could potentially fix many problems.

-- 
  Denise Draper
  ***@***.***

danielmk:
concentratedsolar passes testing on my end without any changes.

denised:
The list in the issue was correct at that time, but probably isn't very relevant now.  The current test results are what to go by.

-- 
  Denise Draper
  ***@***.***

PR #448, opened 2021-09-16
Potential fix for multiple test failures related to HelperTables

danielmk:
This PR relates to #317 where we've been trying to find a fix for multiple, possibly related testing failures.

I found that for the composting solution, adding copy_pds_datapoint=False to the definition of `HelperTables` solves the testing failure.

I also believe that the same change should be bikeinfrastructure & buildingautomation
It fixes the current cause for testing failure but then another testing failure occurs.

I will check how this small change affects the other problematic solutions discussed in the issue and report as soon as I think this PR is ready for merge.

### DIFF
--- a/solution/bikeinfrastructure/__init__.py
+++ b/solution/bikeinfrastructure/__init__.py
@@ -179,7 +179,8 @@ class Scenario(scenario.RRSScenario):
             use_first_pds_datapoint_main=False,
             copy_pds_to_ref=False, copy_ref_datapoint=False,
             pds_adoption_trend_per_region=pds_adoption_trend_per_region,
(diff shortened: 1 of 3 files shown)


denised:
Good to see progress!

danielmk:
This specific `copy_pds_datapoint=False` fix solved testing failures for composting and walkablecities.

In bikeinfrastructure it solves an issue but testing still fails because of another unrelated problem.

In the other scenarios I tried it on it either changes nothing or makes things worse.

I suggest merging this PR since the other solutions will probably require different types of fixes.

denised (approved):

denised:
Got it --- thanks!

PR #456, opened 2021-09-21
Rename customadoption scenario variables (which are not Scenarios) to be custom_pds_scenarios. No-op.

johnpaulalex:
Hopefully this will make their type clearer. No-op.
(Verified by running commercialglass tests).


denised (approved):
I like it!

denised:
Serves me right for not running the entire test suite first....
@johnpaulalex : there are lots of places that use that "scenarios" field.
You are welcome to re-contribute this change if you want to go fix it more globally.

PR #458, opened 2021-09-21
minor progress on integrations

denised:
First two commits update and clarify workbook for waste integration

The last commit moves the initialization of tams and adoptions inside the __init__ function so they will be late-binding.  It isn't much of an issue now, because we have to restart python whenever integrations make model changes.  But there could be a situation where someone, say, does some work with a solution, then continues an in-progress adoption, and they would get the wrong files because the non-integration files would already have been loaded when the solution module was loaded.   It will also matter if/when we are able to do cache busting.


danielmk (approved):
Deferring file loading inside the actual `__init__` makes total sense to me anyway and corresponds more to my understanding of "initializing" a solution.

PR #461, opened 2021-09-23
Code to pull in VMA data and some evaluated electrical demand results

denised:
I made some changes to VMAs on some previous commits, and now am using that code to access VMAs from an integration.
Also an example of letting Scenario.py do some calculation that integration needs.


denised:
This content got pulled into my next PR, so I'm deleting this one.

PR #462, opened 2021-09-23
grid mix initialization

denised:
This is a little different than other integrations so far in that there is a definite place where researchers intervene in the process.  So this is my first approach for setting an initial value for something with the assumption that the user will alter it before proceeding.  (the Jupyter notebook doesn't convey this yet.)   The value they set is immediately saved to an integration-version of the file, so it will be re-used if the user exits and restarts the integration, but can also be cleaned up with the rest if they want to start over.


denised:
Yikes, I thought switching to a new branch would cause only the new commits to be included in the PR, but I guess it doesn't work that way.  I will kill the old PR to avoid conflicts.

danielmk:
I don't understand where and how the researcher is supposed to intervene? If intervening is required, it might be a better idea to raise an error if input is lacking instead of supplying a default. For the development stage a default is probably okay though.

denised:
That's a good idea.  Instead of initializing with a default projected value, start with "You can't proceed until you have set the current grid mix, and here's how to get a default projection".

-- 
  Denise Draper
  ***@***.***

denised:
Updates: implemented the suggested interaction if grid mix is missing;  added testmode data and loading for some cases, and calculating updated PDS Energy TAM (though I don't write it out yet).
Also added a utility to factory.py to translate commonly used titles of solutions to their module name.

PR #463, opened 2021-09-23
Fix extraction of solution description from Excel - column E not D.

johnpaulalex:
This bug resulted in the bogus extraction of description text consisting only of the word 'Description:' for solutions commercialglass, hcrecycling, hfc_replacement, methaneleak, recycledmetals, recycledplastics, residentialglass, and sustainableclothing. (Those solutions still need a fix in their submitted ac/*.json files).


denised (approved):
Thank you --- embarrassing bug on my part :-)

PR #464, opened 2021-09-24
Extract creation_date from excel, validate, store in ac json.

johnpaulalex:

PR #466, opened 2021-09-24
Fix extraction of RRS solution description from Excel - column E not D. 

johnpaulalex:
Land was similarly fixed in PR #463.


denised:
This PR is not functioning properly --- see comments on #464

PR #467, opened 2021-09-24
Extract creation_date from excel, store in ac json.

johnpaulalex:


denised:
I will accept this PR as is, but I realize that advanced_controls.py also needs to be changed, to accept the new field... otherwise the modified scenarios won't load.  @johnpaulalex, can you add that in to your next PR?

denised (approved):

PR #468, opened 2021-09-25
First look at script to extract data from expected.zip and update the…

johnpaulalex:
… scenario jsons. Does not include the ~300 modified json's.

Note also that there's one expected.zip which needs a slight tweak for this to totally work - there's a space at the end of one of the scenario subdirs.

### DIFF
--- a/tools/copy_expected_to_scenario.py
+++ b/tools/copy_expected_to_scenario.py
@@ -1,11 +1,17 @@
 """Template for extracting data from the expected csv's back into scenario json.
 """
+import datetime
 import glob
 import os
 import json
+import pathlib
+
 import pandas as pd
 import zipfile
 import importlib
+
+from solution import factory
+from tools import solution_xls_extract as sxe
 
 # Note that this timestamp could be one second off from the original Excel because our exported 
 # float precision is a bit too low. We export floats like
@@ -28,16 +34,6 @@ def find_scenario_in_record(df_expected, scenario_name):
             break
     return scenario_start_row
 
-def all_solutions(solution_basedir):
-    """Return all solutions with a solution/ subdir."""
-    solution_pys = glob.glob(os.path.join(solution_basedir, '*', '__init__.py'))
-    solutions = []
-    for solution_py in solution_pys:
-        solution_dirname = os.path.dirname(solution_py)
-        solutions.append(os.path.split(solution_dirname)[-1])
-    return sorted(solutions)
-
-
 def copy_expected_to_ac_json(solution_basedir, solution_name):
     solution_dir = os.path.join(solution_basedir, solution_name)
     expected_filename = os.path.join(solution_dir, 'tests', 'expected.zip')
@@ -46,12 +42,14 @@ def copy_expected_to_ac_json(solution_basedir, solution_name):
     with zipfile.ZipFile(expected_filename) as zf:
         for ac_json in ac_jsons:
             print(ac_json)
-            jsonfile = Path(ac_json).resolve()
+            jsonfile = pathlib.Path(ac_json).resolve()
             d = json.loads( jsonfile.read_text(encoding='utf-8') )
             scenario_name = d['name']
 
             # Read exported tab for the scenario from expected.zip. Note that the ScenarioRecord tab is especially
-            # confusing: there's a separate exported file for each scenario, and each one holds _all_ the scenarios.
+            # confusing: there's a separate exported file for each scenario, and each one holds _all_ the scenarios. We conservatively read only scenario x's
+            # rows from scenario x's ScenarioRecord in case the exports happened
+            # at different times.
             sr_file = zf.open(scenario_name + "/" + 'ScenarioRecord')
             df_expected = pd.read_csv(sr_file, header=None, na_values=['#REF!', '#DIV/0!', '#VALUE!', '(N/A)'])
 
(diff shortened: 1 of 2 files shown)


tools/copy_expected_to_scenario.py
```diff
@@ -0,0 +1,84 @@
...
+    with zipfile.ZipFile(expected_filename) as zf:
+        for ac_json in ac_jsons:
+            print(ac_json)
+            jsonfile = Path(ac_json).resolve()
+            d = json.loads( jsonfile.read_text(encoding='utf-8') )
+            scenario_name = d['name']
+
+            # Read exported tab for the scenario from expected.zip. Note that the ScenarioRecord tab is especially
+            # confusing: there's a separate exported file for each scenario, and each one holds _all_ the scenarios.
+            sr_file = zf.open(scenario_name + "/" + 'ScenarioRecord')
```
denised, line 55: And in fact if the information you need is _on_ the ScenarioRecord tab, you only need to read that tab once --- this tab does not change from scenario to scenario.
For other situations, you might need to do something different.

tools/copy_expected_to_scenario.py
```diff
@@ -0,0 +1,84 @@
...
+    name_header_col = df_expected[:][3]
+    names_col = df_expected[:][4]
+    scenario_start_row = None
+    for row_idx, val in enumerate(name_header_col):
+        if val == 'Name of Scenario:' and names_col[row_idx] == scenario_name:
+            scenario_start_row = row_idx
+            break
+    return scenario_start_row
+
+def all_solutions(solution_basedir):
```
denised, line 31: There's a function for this already in factory.py.  Unless there's a reason to avoid that dependency, you might as well use it.

tools/copy_expected_to_scenario.py
```diff
@@ -0,0 +1,84 @@
...
+                creation_date = from_excel_timestamp(creation_date_float)
+            d['creation_date'] = creation_date.strftime(date_format)
+            
+            # Rewrite the json.            
+            sxe.write_scenario(jsonfile, d)
+    return ac_jsons
+
+# Run over all solutions. 
+solution_basedir = f'/Users/jpalex/dd/solutions/solution'  # Modify this as needed.
+all_jsons_modified = []
```
denised, line 78: Our standard practice is to navigate from wherever the file is to where you need to look via the __path__ magic variable.  So you don't need to hardcode a path here;  in pathlib terms it is Path(__file__).parent[1]/"solution"

tools/copy_expected_to_scenario.py
```diff
@@ -0,0 +1,84 @@
...
+            else:
+                creation_date_float = float(creation_date_str)
+                creation_date = from_excel_timestamp(creation_date_float)
+            d['creation_date'] = creation_date.strftime(date_format)
+            
+            # Rewrite the json.            
+            sxe.write_scenario(jsonfile, d)
+    return ac_jsons
+
+# Run over all solutions.
```
denised, line 76: Even though this is an example script, I'd prefer to see "if __name__ == '__main__'" at the top of the script part.

tools/copy_expected_to_scenario.py
```diff
@@ -0,0 +1,84 @@
...
+            creation_date_str = df_expected.iloc[scenario_start_row,1]
+            if ':' in creation_date_str and '-' in creation_date_str:
+                creation_date = datetime.datetime.strptime(creation_date_str, date_format)
+            else:
+                creation_date_float = float(creation_date_str)
+                creation_date = from_excel_timestamp(creation_date_float)
+            d['creation_date'] = creation_date.strftime(date_format)
+            
+            # Rewrite the json.            
+            sxe.write_scenario(jsonfile, d)
```
denised, line 73: Missing dependency sxe?  Or am I not seeing it somewhere...

denised (changes requested):
A couple of nits to fix, but I like it.
I think for educational purposes, it would be good to demonstrate reading the ScenarioRecord once only, but also leave a function that can find the file in the zip for a (scenario, tab) combination

johnpaulalex:
Thanks Denise. All the comments SGTM except reading ScenarioRecord once.  I had considered that but I thought it's _not_ guaranteed to be the same across scenarios due to the updates we plan to support:
   * export scenarios 1, 2, 3 - each has a ScenarioRecord with all 3 scenarios.
   * later, export scenario 4.  It has a ScenarioRecord with all 4, the others don't.
My conclusion was that the only trustworthy version of scenario x's ScenarioRecord is in scenario x's directory. This also seems consistent with the general idea that scenario x's directory holds all scenario x data; the other scenario rows are just noise.

...or can we assume that every export will always update all scenarios?  I guess I thought earlier scenarios might sometimes become unexportable or you simply might want to keep them frozen.  Or you might run the export method and want to revert some of the changes it made to existing files, for some other reason...

denised:
That's a good caution to take, but in this case you can assume that ScenarioRecord is fixed.  We create expected.zip all at once, and we dump everything from the Excel into expected.zip

-- 
  Denise Draper
  ***@***.***

denised:
Actually now that I think about it a bit more, I think you are right.   We create expected.zip all at once now, but we might do that differently later.

-- 
  Denise Draper
  ***@***.***

tools/copy_expected_to_scenario.py
johnpaulalex: SG but when I tried 'from solution import factory' and then ran 'python tools/copy_expected_to_scenario.py', it complained that it couldn't find the 'solution' module.  I feel like I'm missing something basic here.  I thought there might need to be a solution/__init__.py to signal that it's a module directory, but I see you import this module in other files already...

denised: It took me forever to figure this out...
When you execute a script like so:
```
python foo/bar.py
```
Python executes that script *outside* the package hierarchy, and so the code will only find the imports if you have added the solutions directory to your Python path.
When you execute a script like so:
```
python -m foo.bar
```
it operates within the python package hierarchy and relative imports work.

One way I get around this is by working in Jupyter Notebook a lot, and my notebooks always begin by adding solutions to the python path.

(And I only figured this out recently: the reason there is a file `rrs.py` in the tools directory is because I couldn't figure out how to get the relative path working and I just got frustrated and copied the file :-)

tools/copy_expected_to_scenario.py
johnpaulalex: as discussed offline, leaving logic as-is. I added a comment about a contract for ScenarioRecord: only trust the data about scenario x when reading scenario x's ScenarioRecord.

tools/copy_expected_to_scenario.py
johnpaulalex: wow that is really good to know, thanks! All good now.

tools/copy_expected_to_scenario.py
johnpaulalex: done

tools/copy_expected_to_scenario.py
johnpaulalex: fixed

denised (approved):

PR #469, opened 2021-09-25
Add creation_date to AdvancedControls python object, to correspond with its addition to the .json files.

johnpaulalex:
Can't verify. I ran 'python -m pytest solution/commercialglass'
and it said
___ ERROR collecting solution/commercialglass/tests/test_commercialglass.py ____
solution/commercialglass/tests/test_commercialglass.py:6: in <module>
    from tools import expected_result_tester
tools/expected_result_tester.py:16: in <module>
    from model import scenario
model/scenario.py:15: in <module>
    from solution import factory
solution/factory.py:25: in <module>
    def load_scenario(solution, scenario=None) -> scenario.Scenario :
E   AttributeError: partially initialized module 'model.scenario' has no attribute 'Scenario' (most likely due to a circular import)

...and that happens on a clean branch too, so there's some other issue that needs fixing before we can confirm this PR is ok.

### DIFF
--- a/model/advanced_controls.py
+++ b/model/advanced_controls.py
@@ -53,6 +53,9 @@ class AdvancedControls:
     # description: freeform text describing the construction or intention for this scenario.
     description: str = None
 
+    # Creation date of this scenario, in %Y-%m-%d %H:%M:%S format.
+    creation_date: str = None
+
     # js: JSON this AdvancedControls object was created from (if any)
     # jsfile: the filename containing the JSON (if any)
     js: str = None


denised:
Wow, that is obscure.  It happens to me too, but only when I try to run pytest on solution or as you did, a specific solution.  Running just 'pytest' (on the whole directory) works fine.  I can see the problem is the type hint I added to load_scenario recently, so I will remove that asap, since it is not necessary.

Someday maybe I'll figure out why this happened at a deeper level.

-- 
  Denise Draper
  ***@***.***

denised:
Pull the most recent commit and you should be good to go.

-- 
  Denise Draper
  ***@***.***

DentonGentry:
My suspicion would be the metaclass cache https://github.com/ProjectDrawdown/solutions/blob/develop/model/metaclass_cache.py, which is a sizable performance win but can result in caching something across pytest cases. In this case, I guess it would be caching something which is fully initialized where the individual test doesn't fully initialize it? If you put your type hint back, but remove the metaclass cache, and the error goes away that would be pretty telling.

The metaclass cache had been pretty well debugged previously, but still seems plausible.

denised:
@DentonGentry Interesting thought --- I'll give it a shot.

johnpaulalex:
The issue is the circular dependency, no? scenario and factory import each other. Seems like factory is a pretty small class and the two modules don't have a lot of dependencies between them; there are probably a few ways to untangle it.

I would guess that they load successfully if you import scenario first, but not if factory is imported first. So maybe running all the tests coincidentally imports scenario first? Seems like factory needs scenario on startup (to dereference the type hint) whereas scenario doesn't need factory (it calls it in a method, which won't try to actually dereference until it's called).

johnpaulalex:
Ok I repulled and I think this PR is ready to go - PTAL.

denised (approved):

PR #481, opened 2021-10-05
Start of AdvancedControl refactoring to add global support + more

denised:
What is here:
* Create a new code directory called meta_model that has things that manage our data but are not domain specific.   Things like caching, integration support, and so forth.  When we do the data mediator, it will go here.
* Split the implementation of AdvancedControls in two: ParameterCollection is in meta_model and has the code that handles both global variables and VMA substitution. 
* Draft implementation of ParameterCollection, which simplifies (and drops some features from) VMA substitution, adds global variables, adds support for "additional fields", and makes individual fields able to declare verifiers/initializers.

Things I need to do before we can even run the code:
* Complete the implementation of AdvancedControls as inheriting from ParameterCollection
* Fix up all the package references  between model and meta_model 
* Fix up the initialization of VMAs so that the solutions define which VMAs go with which parameters.
* Complete the code path for reading and writing ac objects.

Beyond that:
 * Implement pytest fixtures to extract certain values from expected.zip and set them explicitly.

Other things I want to do as well:
 * Move the list of VMA declarations in `__init__.py` into a json file, just like the adoptions and tams.
 * Move the "fixed summaries" of VMAs into the VMA metadata
 * Rename AdvancedControls --> ScenarioParameters or something like that.
 * Unify treatment of parameters between old-style models and the new ocean models.


denised:
I'm going to be leaving this work for now to work on other things.
What has been finished:

- Package references for all meta_model packages have been fixed across the code base
- Advanced Controls has been rewritten to use the new ParameterClass format.  I also did some tidying up, but not all.
- Advanced Controls has been split into AdvancedControls and LandAC, where LandAC inherits from AC and has all the Land-only fields
- These have been tested to work in isolation, but not together with the rest of the system.

These items still remain to be done:

- Fix up the initialization of VMAs so that the solutions define which VMAs go with which parameters.
- Complete the code path for reading and writing ac objects.
- Implement pytest fixtures to extract global values from expected.zip and set them explicitly before beginning a test.

PR #510, opened 2021-11-12
Rewrite of vma_xls_extract.py

denised:
While extracting a solution I had an issue with VMAs, and found myself getting lost in the vma extraction code (not for the first time).  So this is a rewrite, aimed at readability.  It has almost exactly the same functionality as the original, with some simplifications.  Its a little slower, because it does multiple passes over the data.  On the plus side, it doesn't have any hardwired dependencies on tab names, rows or columns.   But mostly, I believe it is a lot easier to read, so when issues arise it should be much easier to understand what is causing them.

I also revised some functionality in util.py, and created a single "turn a string into a filename" function that is shared between both extraction modules.


denised:
@johnpaulalex For some reason I can't select you as a reviewer, but I would appreciate it if you would take a look at this code.

tools/vma_xls_extract.py
```diff
@@ -1,483 +1,281 @@
...
-            if sheet.cell(r, tools.util.co("S")).value == 'Use weight?':
-                use_weight = tools.util.convert_bool(sheet.cell(r+1, tools.util.co("S")).value)
-                break
+def get_vma_sheet(wb):
+    """Return the preferred VMA worksheet in this workbook"""
+    if "Variable Meta-analysis-DD" in wb.sheetnames:
+        return "Variable Meta-analysis-DD"
+    if "Variable Meta-analysis" in wb.sheetnames:
+        return "Variable Meta-analysis"
+    return None
```
johnpaulalex, line 33: callers don't seem ready for None. Maybe throw an exception (with the two desired names + the actual wb.sheetnames in the err str, so callers can see what kind of mismatch happened)

tools/vma_xls_extract.py
```diff
@@ -1,483 +1,281 @@
...
+    tableinfo = find_vma_tables(ws, start_row=start_row, start_column=start_column)
+    results = []
+    for loc in tableinfo:
+        d = get_vma_table_info(ws,*loc)
+        if d is not None:
+            df = get_vma_table_data(ws,*loc)
+            if df is not None:
+                d['data'] = df
+                results.append(d)
+    return results
```
johnpaulalex, line 49: (paranoia) should this throw if no results are found?

johnpaulalex, line 49: It's a broader problem in the codebase, but it's hard to figure out what the data type of 'results' is. Is there anything for it here?

ps have you considered https://github.com/google/pytype

tools/vma_xls_extract.py
```diff
@@ -1,483 +1,281 @@
...
+        vma_d['data'].to_csv(outputdir/filename, encoding='utf-8', index=False)
+        dcopy = vma_d.copy()
+        del(dcopy['name'])
+        del(dcopy['data'])
+        dcopy['filename'] = filename
+        dirinfo[vma_name] = dcopy
+    (outputdir/"vma_sources.json").write_text(json.dumps(dirinfo,indent=2))   
+
+
+def find_vma_tables(ws, start_row, start_column):
```
johnpaulalex, line 72: at fiirst glance this seems like the same idea of extract_vmas()...

johnpaulalex: on second glance extract_vmas() does some additional postprocessing on this method? Maybe doc the difference somewhere, and/or give the methods a similar name, e.g.
find_vma_tables and find_vma_tables_and_whatever_that_postprocessing_is_doing()

tools/vma_xls_extract.py
```diff
@@ -1,483 +1,281 @@
...
+    
+    # We use the presence of the "SOURCE ID: Author/Org, Date, Info" information to identify
+    # the start of the table, and the presence of "**add calc above" in the "Conversion
+    # calculation**" column to identify the bottom.  We also assume that all the VMAs on this
+    # sheet will have Exclude Data? in the same column, so we don't need to keep searching for
+    # it.  Note this returns all VMAs, even if they are empty.
+
+    firstrow = find_in_column(ws, start_column, STANDARD_COLUMNS[0], start_row)
+    if firstrow is None:
+        raise ValueError("Unable to find any VMA tables on Excel page")
```
johnpaulalex, line 90: maybe include the worksheet name in all these errors?

johnpaulalex, line 90: maybe give some signal on how the code decided this - maybe include STANDARD_COLUMNS[0] in the error?

tools/vma_xls_extract.py
```diff
@@ -1,483 +1,281 @@
...
-        bound_correction_cols = ["X", "Y"]
-        for r in range(last_row, last_row + 100):
-            for col in bound_correction_cols:
-                if sheet.cell(r, tools.util.co(col)).value == 'Low Correction?':
-                    bound_correction = tools.util.convert_bool(
-                        sheet.cell(r+1, tools.util.co(col)).value)
-                    break
-        if bound_correction is None:
-            raise ValueError(f"No 'Low Correction?' cell found for VMA between {last_row} and {last_row+100} in columns {bound_correction_cols}.")
+            raise ValueError("Unable to find last column of VMA table at " + str(firstrow))
```
johnpaulalex, line 99: mention 'Exclude Data' in here

tools/vma_xls_extract.py
```diff
@@ -1,483 +1,281 @@
...
+
+    result = []
+    while firstrow is not None:
+        # To minimize the possibility of mistakenly finding a range that spans multiple tables,
+        # we actually look for the next table first, then look for the end of this table before that.
+        next_firstrow = find_in_column(ws, start_column, STANDARD_COLUMNS[0], start_row=firstrow+1)
+        lastrow = find_in_column(ws, checkcol, "**Add calc above", start_row=firstrow+1, end_row=next_firstrow)
+        if lastrow is None:
+            raise ValueError("Unable to find end of VMA table at " + str(firstrow))
+        result.append((firstrow, start_column, lastrow-1, lastcol))
```
johnpaulalex, line 113: maybe opening a bigger can of worms but how about making this a data object (so then you can document what type this is as you pass it around), e.g. python attrs classes.

johnpaulalex, line 113: maybe opening a bigger can of worms but how about making this a data object (so then you can document what type this is as you pass it around), e.g. python attrs classes.

johnpaulalex (approved):
I like this new VMA code, I understand what it's doing :) Is it tested? 

Mainly my comments are around making the exceptions as informative as possible. 

And I guess a more general comment about opaque data types - I guess they're often black-box json dicts or dataframes.

tools/vma_xls_extract.py
denised: I will make it an error, I think that is a good call.
(I don't like the idea of listing available sheetnames, though, as there may be dozens, and that kind of verbosity obscures more than it illuminates IMHO.)

tools/vma_xls_extract.py
denised: I will add that where it is called in solution extract, rather than here.

denised: I don't feel so bad about untyped structs in bits of code that are designed to work with each other and not expected to be widely resused or passed outside limited scope of code.   But its certainly a sliding scale.   Pytype is new(s) to me and looks very interesting!

tools/vma_xls_extract.py
denised: I'll improve some of the naming and docstrings.

tools/vma_xls_extract.py
denised: My thinking here is that the code is magic enough that you are going to have to read it to figure out what is going on anyway --- this is in large part the motivation for the rewrite, that when you do run into difficulties, the code is readable enough that you can understand what happened.  I don't think I can construct an error message that would be clear enough to avoid this.

denised: Done.

denised:
Thank you @johnpaulalex for the very helpful review.  I have included some lightweight pytest tests for the new code, but the more in-depth testing I did was a side-by-side comparison of the old and new implementations on all the recently imported solutions.  (Which doesn't yet include any land models, so I will make a point of doing one of those soon.)

...And I just now spent a a few hours adding more testing of error/edge cases... which netted four bugs, so here's to testing :-)

PR #512, opened 2021-11-18
smartglass solution

sutjin:
Deep test showing error but only on less than 10% of the rows for Unit Adoption Calculations


denised (changes requested):
Please remove all but the most recent three scenarios from the ac folder.

denised:
I took a look at this one, because Smartglass didn't use to have an error.  I figured out what was different.  It is a minor, somewhat inconsequential issue, but for the purpose of (a) keeping the tests passing and (b) learning a bit more about how to deal with weirdness in the code, I'm going to push it back to you and let you fix it.

The Unit Adoption table that is kicking up errors is simply a restatement of the TAM.  The errors are only in column two, which is to say, the OECD region.  Looking at the TAM page in Excel, it becomes clear that the same data is in the World table and in the OECD table --- this is strange ,but comments in the source names lead you to a different tab that explains what they are doing.  (Still a questionable approach, IMO, but that is a different discussion.)  If you look at the Unit Adoption Table in the Excel, you see that the World column and the OECD column have exactly the same output values.  Hence they are being calculated in the same way.  The python code is clearly not doing the same calculation -- the question is why, and how can it be fixed.  It turns out this is not something that the extractor currently even looks for, but it is in bright red on the TAM configuration sheet.  You can override the TAM configuration values to do the right thing.

sutjin:
@denised - noted, I will remove the additional scenario and review that problem

denised (approved):

PR #515, opened 2021-11-28
bug fixes & improvements to extractor

denised:
Big improvement: change code for helper tables quirks detection to look directly at excel formulas instead of guessing from data values.
Improvement: put main code inside try/finally to close files in case of error; allows to restart extractor without having to restart python.
Bug fix: format of string @line 862
Bug fix: read multipliers as floats not ints @line 1306


denised:
@sutjin I see you've got a work item for improvements to the extractor --- maybe you can review this one :-)

